PR #1, opened 2021-11-18
hwc: Pointer device enable natural scrolling by default

atmouse-:
been tested OK on mainline device


electrikjesus (approved):
This works on my end

PR #2, opened 2022-01-05
Fix Bug : key handle error

haz102:
Fixed : key event doesn't fired clearly due to dirty buffers.

### DIFF
--- a/hwcomposer/wayland-hwc.cpp
+++ b/hwcomposer/wayland-hwc.cpp
@@ -566,7 +566,7 @@ keyboard_handle_key(void *data, struct wl_keyboard *,
     }
     ADD_EVENT(EV_KEY, key, state);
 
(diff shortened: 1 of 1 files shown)


haz102:
@erfanoabdi kindly request, please review this patch.

electrikjesus:
How can we recreate the issue to verify?

electrikjesus:
And some feedback from our discussion: 
Erfan Abdi:
better fix would be using 1 instead of 6 on struct input_event event[6];
and we should do the same for other inputs

1 for keyboard_handle_key
5 for pointer_handle_motion
2 for pointer_handle_button
..

(as many times ADD_EVENT is called on each function)

erfanoabdi:
https://github.com/waydroid/android_hardware_waydroid/commit/bc85e3c88798f6bbecc8fef0fc4d9370f679d83f

Should do the same thing

PR #6, opened 2022-06-18
wayland-hwc: Drop dummy window hack & QtMir integration improvements

fredldotme:
- The 1x1 dummy buffer causes QtMir to only draw that buffer, not the
  attached surfaces. Remove the hack.
- Allow tracking window spawn state through Android properties.
- Various improvements to integration into Mir & QtMir

### DIFF
--- a/hwcomposer/hwcomposer.cpp
+++ b/hwcomposer/hwcomposer.cpp
@@ -554,9 +554,12 @@ static int hwc_set(struct hwc_composer_device_1* dev,size_t numDisplays,
                     property_set("waydroid.open_windows", std::to_string(pdev->windows.size()).c_str());
                 }
                 window = pdev->windows[single_layer_tid];
+
                 // Window is closed, don't bother
-                if (!window->isActive)
+                if (window && !window->isActive) {
+                    free(window);
                     window = NULL;
+                }
             }
         } else {
             // Create windows based on Task ID in layer name
@@ -585,8 +588,10 @@ static int hwc_set(struct hwc_composer_device_1* dev,size_t numDisplays,
                 }
 
                 // Window is closed, don't bother
-                if (window && !window->isActive)
+                if (window && !window->isActive) {
+                    free(window);
                     window = NULL;
+                }
             }
         }
 
--- a/hwcomposer/hwcomposer.cpp
+++ b/hwcomposer/hwcomposer.cpp
@@ -583,6 +583,10 @@ static int hwc_set(struct hwc_composer_device_1* dev,size_t numDisplays,
                     if (pdev->windows.find(layer_tid) != pdev->windows.end())
                         window = pdev->windows[layer_tid];
                 }
+
+                // Window is closed, don't bother
+                if (window && !window->isActive)
+                    window = NULL;
             }
         }
 
--- a/hwcomposer/hwcomposer.cpp
+++ b/hwcomposer/hwcomposer.cpp
@@ -1006,6 +1006,9 @@ static int hwc_open(const struct hw_module_t* module, const char* name,
     if (!property_get_bool("persist.waydroid.cursor_on_subsurface", false))
         pdev->display->cursor_surface =
             wl_compositor_create_surface(pdev->display->compositor);
+    else
+        pdev->display->cursor_surface = nullptr;
+
     if (!pdev->display->height) {
         pdev->display->waiting_for_data = true;
         pthread_cond_timedwait(&pdev->display->data_available_cond,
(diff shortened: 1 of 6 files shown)


fredldotme:
The window spawn state tracking is necessary for this feature: https://github.com/waydroid/waydroid/pull/418

fredldotme:
The question remains whether the EGL window change is actually necessary.
I'd gladly rip it out considering surfaces work properly anyway and the 1x1 buffer hack for positioning windows is illegal in Wayland anyway.
It's job of the compositor to place (and remember) positions of windows when spawning it.

dos1:
Wouldn't either setting xdg-surface's window geometry or using viewporter to extend the single pixel buffer into desired size be enough there? I'm not sure what wayland-egl brings to the table.

fredldotme:
The wayland-egl stuff was an experiment, sadly failed.

Problem with the single pixel solution is that it might work for one or two compositors, but some might expose weird behavior, like QtMir does where the 1px buffer blocks view to the (sub)surfaces.

The reasoning behind it is also a little problematic as apps being able to decide where they want to place themselves is precisely something that Wayland tries to counteract. Nothing guarantees us that GNOME Shell/Mutter or some other major compositor kills that happy accident in an update due to trying to stay true to the protocol.

dos1:
Well, the xdg-surface needs some buffer because otherwise it will stay unmapped. In subsurface mode HWC essentially offloads compositing to the Wayland compositor via subsurfaces, and since nothing on Android side guarantees that the first layer always starts at 0x0 you need to have a dummy buffer regardless of whether you're trying to hackily position your windows or not. [edit] <strike>Well, I guess you could use buffer offset and place the first layer there, but I think a dummy toplevel buffer can be still valuable as I'd like to allow subsurfaces to work in single-window mode as well and that buffer could then provide a solid black background below all the layers.</strike> scratch that, I was right at first, buffer offset wouldn't help there :P

Does this commit - d8ad35e0c7deff363645f6e2b3a3c74c41144e87 - fix the problem with QtMir? (it's in a branch I'm in the middle of cleaning up right now)

dos1:
I have now posted the window geometry patch as a pull request (#18) and also posted a change that makes use of the dummy buffer to provide window background (#12).

It will still be desirable to position each window's layers independently of global coordinate space, but that will require some more involved changes. Your PR didn't really change anything in that regard - it just replicated pretty much the same hack by creating the buffer in a different way. I'll be interested in whether either #12 or #18 (or maybe both combined) fix your QtMir issues though, and from what I can see this PR still has interesting and useful changes in it that could be split out.

hwcomposer/wayland-hwc.cpp
```diff
@@ -412,10 +415,15 @@ destroy_window(struct window *window, bool keep)
         window->isActive = false;
     else
         free(window);
+
+    if (!window->appID.empty()) {
```
dos1, line 419: Use after free.

hwcomposer/hwcomposer.cpp
```diff
@@ -533,26 +533,33 @@ static int hwc_set(struct hwc_composer_device_1* dev,size_t numDisplays,
...
-                    pdev->windows[single_layer_tid] = create_window(pdev->display, pdev->use_subsurface, single_layer_aid, single_layer_tid);
+                    pdev->windows[single_layer_tid] = create_window(pdev->display, pdev->use_subsurface, single_layer_aid, single_layer_tid, width, height);
                     property_set("waydroid.open_windows", std::to_string(pdev->windows.size()).c_str());
                 }
                 window = pdev->windows[single_layer_tid];
+
                 // Window is closed, don't bother
-                if (!window->isActive)
+                if (window && !window->isActive) {
+                    free(window);
```
dos1, line 560: You're not removing the window from the list, which leads to crash. Also, I'd rather use `destroy_window` instead of freeing directly.

hwcomposer/hwcomposer.cpp
```diff
@@ -1006,6 +1006,9 @@ static int hwc_open(const struct hw_module_t* module, const char* name,
     if (!property_get_bool("persist.waydroid.cursor_on_subsurface", false))
         pdev->display->cursor_surface =
             wl_compositor_create_surface(pdev->display->compositor);
+    else
+        pdev->display->cursor_surface = nullptr;
```
dos1, line 1019: I'm not really well-versed in C++, but since the display is allocated with `struct display *display = new struct display();`, this should mean that all its members get value-initialized, which in case of pointers initializes them to `nullptr` already. Is that not the case here?

PR #8, opened 2022-07-04
gralloc_gbm: Always map the full buffer size

aleasto:
Upstream maintainer seems to be MIA, but patch is good.


erfanoabdi:
looks good to me but what does it fix?

aleasto:
> looks good to me but what does it fix?

https://github.com/robherring/gbm_gralloc/issues/22

aleasto:
Merged upstream

PR #9, opened 2022-07-08
Scroll

aleasto:


hwcomposer/wayland-hwc.cpp
```diff
@@ -710,34 +710,49 @@ pointer_handle_axis(void *data, struct wl_pointer *,
     struct display* display = (struct display*)data;
     struct input_event event[2];
     struct timespec rt;
-    unsigned int res, n = 0;
+    unsigned int res, v, n = 0;
+    static const int step = 4000;
```
aleasto: Kinda arbitrary, but i think it should be a decent default. Please test on your touchpads!

aleasto:
Huh, thought this was merged

PR #11, opened 2022-08-16
gralloc_gbm: Add SCANOUT flag to HW_FB and HW_COMPOSER usages

dos1:
Some hardware configurations require the BO to be created with
GBM_BO_USE_SCANOUT in order to make it suitable for display.

This fixes garbled buffer content on Purism Librem 5 and likely
various multi-GPU PC configurations too.

### DIFF
--- a/gralloc/gralloc_gbm.cpp
+++ b/gralloc/gralloc_gbm.cpp
@@ -153,9 +153,9 @@ static unsigned int get_pipe_bind(int usage)
 	if (usage & (GRALLOC_USAGE_HW_RENDER | GRALLOC_USAGE_HW_TEXTURE))
 		bind |= GBM_BO_USE_RENDERING;
 	if (usage & GRALLOC_USAGE_HW_FB)
-		bind |= GBM_BO_USE_RENDERING;
+		bind |= GBM_BO_USE_SCANOUT | GBM_BO_USE_RENDERING;
 	if (usage & GRALLOC_USAGE_HW_COMPOSER)
-		bind |= GBM_BO_USE_RENDERING;
+		bind |= GBM_BO_USE_SCANOUT | GBM_BO_USE_RENDERING;
 
 	return bind;
 }


aleasto:
Can you send this upstream? I'd like a comment from them.

https://github.com/robherring/gbm_gralloc

aleasto:
Also in waydroid 1.3.0 I've made so that only one DRI render node is exposed to the container. You mention this helps with dual GPUs systems: is that still the case?

Quackdoc:
there are still some issues regarding dual gpu, such as if gpu passed through is different from the one the compositor is using, I can routinely and easily replicate issues in such cases, no idea if this will address that but if so it could be really nice, as this is quite common on PCs with nvidia gpu and intel/amd integrated

dos1:
> Can you send this upstream?

Upstream does that already: https://github.com/robherring/gbm_gralloc/blob/master/gralloc_gbm.cpp#L152

> You mention this helps with dual GPUs systems: is that still the case?

Note "likely" - I was just guessing, I don't have any system like that to test on.

However, I'm having second thoughts on this patch. It may still be a good idea to add GBM_BO_USE_SCANOUT so the buffers can end up being placed on hardware planes by Wayland compositor, but as far as I understand all this modifier stuff the fact that the buffer content is garbled without that flag even when it's being GL composited onto the screen means that the modifier information is being lost somewhere on the path. I believe the SCANOUT flag on etnaviv/lcdif will cause the buffer to be linear, so it ends up merely hiding the bug that still exists somewhere...

aleasto:
> > Can you send this upstream?
> 
> Upstream does that already: https://github.com/robherring/gbm_gralloc/blob/master/gralloc_gbm.cpp#L152

I see. So this reverts https://github.com/waydroid/android_hardware_waydroid/commit/3dfded2ca9c1af7e01d450163240ee06e87888af

@erfanoabdi ?

dos1:
I have updated the commit message to make it clear that this only workarounds (as opposed to properly fixing) the issue on Librem 5, removed a mention of multi-GPU systems where I don't really know whether this will help or not, but also mentioned why this is still desirable to do even if nothing had required any workarounds.

aleasto (approved):

aleasto:
In lineage-18.1 this causes the cursor and activity launch animation to misrender (like bad stride or something?)

dos1:
@aleasto Is that on single window mode with subsurfaces disabled?

aleasto:
> @aleasto Is that on single window mode with subsurfaces disabled?

Correct

dos1:
I wonder if this may the exact same issue Librem 5 has without this patch, just manifesting itself in a different way on another hardware with different implicit modifier semantics.

What GPU / mesa driver are you using?

aleasto:
> I wonder if this may the exact same issue Librem 5 has without this patch, just manifesting itself in a different way on another hardware with different implicit modifier semantics.
> 
> What GPU / mesa driver are you using?

Intel with gallium iris.
Indeed in lineage-17.1 we're using the legacy ~~i915~~ i965 driver

dos1:
I have checked the used modifiers on Librem 5 and, at least in lineage-17.1, I can see that while wayland-hwc is correctly passing the modifier it got from gralloc, the buffer content is actually filled up as if it was linear. Changing wayland-hwc to always pass 0 as the modifier to Wayland compositor makes things look fine. This likely means that it's Android's rendering code that disregards the modifier after importing gralloc's handle. At the moment I have no idea where that code sits though - but I'm quite optimistic that figuring this out will be the key to fixing multi-GPU and iris issues.

dos1:
I think I've found it.

When I hardcode the correct modifier here:

```diff
diff --git a/external/mesa3d/src/egl/drivers/dri2/platform_android.c b/external/mesa3d/src/egl/drivers/dri2/platform_android.c
index 099e38699db..a407fdb1f64 100644
--- a/external/mesa3d/src/egl/drivers/dri2/platform_android.c
+++ b/external/mesa3d/src/egl/drivers/dri2/platform_android.c
@@ -372,7 +372,7 @@ droid_create_image_from_native_buffer(_EGLDisplay *disp,
                                       void *priv)
 {
    struct dri2_egl_display *dri2_dpy = dri2_egl_display(disp);
-   struct buffer_info buf_info = {.modifier = DRM_FORMAT_MOD_INVALID};
+   struct buffer_info buf_info = {.modifier = DRM_FORMAT_MOD_VIVANTE_SUPER_TILED};
    unsigned error;
 
    if (!cros_get_buffer_info(disp, buf, &buf_info))

```

things render and display correctly without this PR applied. So it appears that `native_window_buffer_get_buffer_info` needs to somehow learn how to get the modifier information from ANativeWindowBuffer handle.

aleasto:
If I'm reading this correctly, mesa 22.1 that we're using in branch lineage-18.1 already does that.

dos1:
From what I can see, `native_window_buffer_get_buffer_info` still doesn't handle modifiers, but there's `mapper_metadata_get_buffer_info` that potentially can. The question is whether this function ends up being actually used.

Is the manifest for 18.1-based images in android_vendor_waydroid supposed to work already? (in other words, should I spend some CPU hours trying to build it? :))

aleasto:
yep :)

dos1:
FYI, I got:
```
fatal: remove-project element specifies non-existent project: platform/hardware/intel/common/libva
```
then after removing that element the same thing about `LineageOS/android_external_chromium-webview`, and then:
```
fatal: duplicate path external/chromium-webview/patches in /root/android-18/.repo/manifest.xml
```
so I replaced it with:
```
  <remove-project name="LineageOS/android_external_chromium-webview_patches" />
  <remove-project name="LineageOS/android_external_chromium-webview_prebuilt_arm" />
  <remove-project name="LineageOS/android_external_chromium-webview_prebuilt_arm64" />
  <remove-project name="LineageOS/android_external_chromium-webview_prebuilt_x86" />
```
and now it seems to be chugging through downloads.

aleasto:
Sounds like you're using generate-manifest.sh from lineage-17.1 branch

dos1:
Nope, I did:
```
repo init -u https://github.com/LineageOS/android.git -b lineage-18.1 --depth=1
wget -O - https://raw.githubusercontent.com/waydroid/android_vendor_waydroid/lineage-18.1/manifest_scripts/generate-manifest.sh | bash
```

I'm getting many conflicts at `apply-waydroid-patches` too. Something's definitely off there.

aleasto:
Oh. looks like the script is trying to detect the android version from build/make/core/version_defaults.mk but if you haven't synced yet it cannot.

Run `repo sync build/make` before generating the local manifest

dos1:
So I now have a 18.1 image and although I haven't debugged the actual code yet, I can see that it's still rendering linearly to a tiled buffer, so it doesn't look like `mapper_metadata_get_buffer_info` does the job there.

(it does seem to solve some other remaining issues I was having on Librem 5 though, so I'm pretty motivated to get this to work right :D)

aleasto:
Looks like we would need to implement android.hardware.graphics.mapper@4.0
We're currently using android.hardware.graphics.mapper@2.1 default's implementation that just calls into gralloc

aleasto:
Or add android.hardware.graphics.mapper@2.0/2.1 support in mesa if that's feasible

dos1:
I have already implemented the interface from CrOS that Mesa is also able to speak in gralloc_gbm and it works. Having Mapper 4.0 would work too, but that seems much more involved (AFAIUI this would require implementing Gralloc 4 as well?)

aleasto:
Oh you mean `cros_get_buffer_info`? Unfortunately that requires changing mesa to allow that path for gralloc name `!= "CrOS Gralloc"` but I suppose it's an ok change to track downstream

dos1:
Yup, I just added gralloc_gbm's name there too. I think I could also implement reading stuff from gralloc_handle directly in Mesa, but in the end I think simply reusing CrOS interface will be the least invasive to carry downstream.

dos1:
@aleasto Could you test #24 (plus Mesa part) with this commit applied on 18.1? Although with #24, this MR doesn't actually have to be there and I think it may actually be better to leave it out, but it's still worth checking whether #24 fixes the issues you were experiencing on iris with it.

aleasto:
> @aleasto Could you test #24 (plus Mesa part) with this commit applied on 18.1? Although with #24, this MR doesn't actually have to be there and I think it may actually be better to leave it out, but it's still worth checking whether #24 fixes the issues you were experiencing on iris with it.

Nope, same issue

dos1:
Huh, I guess it will not only need modifier information, but also usage flags to be passed into mesa. I think I've seen a separate opcode for that in CrOS.

In any case, #24 alone already fixes garbled display on Librem 5.

PR #12, opened 2022-08-19
hwcomposer: Use dummy buffer as window background in subsurface mode

dos1:
HWC assumes that there's an implicit solid black layer below all the other layers to compose onto. Since we already have a dummy buffer attached to the xdg-toplevel surface, it can be used to provide that background in single window mode, which fixes some apps having transparent background in unexpected places. Thanks to wp_viewporter, the buffer can still be just a single pixel, which gets upscaled to cover the whole window by the compositor.

Since multi window mode relies on the background being transparent, keep it that way there by allowing to specify window background color in create_window.

When the background is transparent, we don't want it to consume any input events.

Similarly, when the background is fully opaque, we know that the whole surface is opaque too, so we can mark it as such to allow the Wayland compositor to optimize its drawing.

Also, don't leak the dummy wl_buffer on window destroy.

### DIFF
--- a/hwcomposer/hwcomposer.cpp
+++ b/hwcomposer/hwcomposer.cpp
@@ -537,14 +537,14 @@ static int hwc_set(struct hwc_composer_device_1* dev,size_t numDisplays,
         if (active_apps == "Waydroid") {
             // Show everything in a single window
             if (pdev->windows.find(active_apps) == pdev->windows.end()) {
-                pdev->windows[active_apps] = create_window(pdev->display, pdev->use_subsurface, active_apps, "0");
+                pdev->windows[active_apps] = create_window(pdev->display, pdev->use_subsurface, active_apps, "0", {0, 0, 0, 255});
                 property_set("waydroid.open_windows", std::to_string(pdev->windows.size()).c_str());
             }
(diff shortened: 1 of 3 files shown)


aleasto (approved):

aleasto:
~~There's some viewport references that i think were meant for #22 that makes this not compile.~~

~~Perhaps prepare a merge request with all the commits rebased to avoid further conflicts~~

This merges and builds clean after #22 that defines the viewporter, but #22 is broken for me.

dos1:
Rebased and fixed. This MR doesn't depend on #22, I just messed it up when splitting things out.

aleasto:
@dos1 looks like the ubports compositor doesn't implement viewporter so this (as well as your other changes involving the viewporter) breaks compositing there...

dos1:
Ah, yeah. Any reasonably modern compositor should already support viewports, but well, UBport's Mir sure doesn't qualify to such title...

Won't be a big deal though, it's just a matter of sprinkling stuff with bunch of ifs. Of course crops will stay broken on such compositors, but at least the background can still be provided by creating a full-sized buffer instead.

PR #13, opened 2022-08-19
hwcomposer: Force single window mode during boot animation

dos1:
This provides feedback to the user that the Android OS is launching.

### DIFF
--- a/hwcomposer/hwcomposer.cpp
+++ b/hwcomposer/hwcomposer.cpp
@@ -454,6 +454,15 @@ static int hwc_set(struct hwc_composer_device_1* dev,size_t numDisplays,
     std::string single_layer_tid;
     std::string single_layer_aid;
 
+    for (size_t l = 0; l < contents->numHwLayers; l++) {
+        std::string layer_name = pdev->display->layer_names[l];
+        if (layer_name.rfind("BootAnimation#", 0) == 0) {
+            // force single window mode during boot animation
+            active_apps = "Waydroid";
+            break;
+        }
+    }
+
     std::scoped_lock lock(pdev->display->windowsMutex);
     if (active_apps == "none") {
         // Clear all open windows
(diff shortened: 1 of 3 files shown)


aleasto:
I believe it was made intentional to not show the boot animation (by looking at how waydroid.py explicitly waits for a boot completed signal before setting `waydroid.active_apps`).

I would also prefer to show the boot animation but i think it's best done by setprop in `waydroid show-full-ui`

dos1:
My main motivation was to make it also visible when launching a single app without show-full-ui, but I guess this could also be handled by waydroid.py (although I haven't looked there at all yet so not 100% sure).

aleasto:
So I went further and actually made the boot animation window the very first window that gets created, removing the _calibration_ window that was used to get the maximized width,height from `configure` events.

I like this, but the concern I still have is that `waydroid session start` used to be a backgroud operation one might add to their startup scripts, while now it would create a window with the boot animation which will then be closed once the boot finishes.

aleasto:
~Oh also Ubuntu Touch doesn't quite like when a process opens and closes windows at will, and this change seems to make it worse in the sense that the Waydroid window doesn't pop back up after the boot animation window closes~
Actually without https://github.com/waydroid/android_vendor_waydroid/pull/18 it's mostly ok

Fuseteam:
> ~Oh also Ubuntu Touch doesn't quite like when a process opens and closes windows at will, and this change seems to make it worse in the sense that the Waydroid window doesn't pop back up after the boot animation window closes~ Actually without [waydroid/android_vendor_waydroid#18](https://github.com/waydroid/android_vendor_waydroid/pull/18) it's mostly ok

This is mostly due to it creating and closing when launching a single app, there was always an empty desktop icon in the left bar without https://github.com/waydroid/android_vendor_waydroid/pull/18

PR #14, opened 2022-08-19
hwcomposer: Invalidate whole buffer cache on HWC_GEOMETRY_CHANGED

dos1:
HWC contract guarantees that a layer buffer can be reused, but only if it didn't change on two consecutive prepare calls *and* the HWC_GEOMETRY_CHANGED flag is not set. With the way buffer cache worked so far, it was possible for an old buffer to get reused after HWC_GEOMETRY_CHANGED was set if it wasn't present in the layer list that had that flag set.

This fixes flickering in subsurface mode.

### DIFF
--- a/hwcomposer/hwcomposer.cpp
+++ b/hwcomposer/hwcomposer.cpp
@@ -135,25 +135,18 @@ static struct buffer *get_wl_buffer(struct waydroid_hwc_composer_device_1 *pdev,
     int height = layer->displayFrame.bottom - layer->displayFrame.top;
     auto it = pdev->display->buffer_map.find(layer->handle);
     if (it != pdev->display->buffer_map.end()) {
-        if (!pdev->display->geo_changed) {
-            if (it->second->isShm) {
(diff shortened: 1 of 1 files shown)


aleasto (approved):

aleasto:
> HWC contract guarantees that a layer buffer can be reused, but only if it didn't change on two consecutive prepare calls and the HWC_GEOMETRY_CHANGED flag is not set

Is this what you're citing?
https://android.googlesource.com/platform/hardware/libhardware/+/master/include/hardware/hwcomposer.h#140

From that description I only get that If HWC_GEOMETRY_CHANGED is set then the **contents** of the buffer might have changed, but we don't care since we always call wl_surface_damage_buffer anyways..?
I'm not sure why reimporting the same prime_fd into wayland would have any effect with respect to "flickering in subsurface mode". In facts i don't see the reason to ever invalidate this cache (other than it might get big over time)?

dos1:
FDs can get reused, and without invalidating you end up using an old and completely different buffer instead of importing the new one in such case.

PR #17, opened 2022-08-19
hwcomposer: Switch focused task on keyboard focus instead of ACTIVATED state

dos1:
XDG_TOPLEVEL_STATE_ACTIVATED cannot be used to track focus. It's only meant to provide visual clues, and some compositors may decide to make windows appear activated even if they're not focused (for example, to make them look better in expose mode or activity switchers). Use keyboard focus instead.

Marked as draft because I haven't had a chance to test it with regular desktopy compositor yet and where I did test it I have issues with focus regardless of this patch's presence - so some external testing will be appreciated.


dos1:
> so some external testing will be appreciated

FWIW there's no need to test it, it's clearly still broken and needs more work.

dos1:
Fixed to actually work now:)

PR #18, opened 2022-08-19
hwcomposer: Set window geometry

dos1:
This should help some compositors figure out the correct window size regardless of subsurface placement.

### DIFF
--- a/hwcomposer/wayland-hwc.cpp
+++ b/hwcomposer/wayland-hwc.cpp
@@ -452,14 +452,10 @@ create_window(struct display *display, bool with_dummy, std::string appID, std::
         if (appID != "Waydroid")
             appID = "waydroid." + appID;
         xdg_toplevel_set_app_id(window->xdg_toplevel, appID.c_str());
-        wl_surface_commit(window->surface);
 
         if (display->isWinResSet)
             xdg_surface_set_window_geometry(window->xdg_surface, 0, 0, display->width / display->scale, display->height / display->scale);
 
-        /* Here we retrieve objects if executed without immed, or error */
-        wl_display_roundtrip(display->display);
-        wl_surface_commit(window->surface);
     } else if (display->shell) {
         window->shell_surface =
             wl_shell_get_shell_surface(display->shell, window->surface);
@@ -475,12 +471,6 @@ create_window(struct display *display, bool with_dummy, std::string appID, std::
                                       { wl_shell_surface_set_title(window->shell_surface, value.c_str()); });
         else
             wl_shell_surface_set_title(window->shell_surface, appID.c_str());
-
-        wl_surface_commit(window->surface);
-
-        /* Here we retrieve objects if executed without immed, or error */
-        wl_display_roundtrip(display->display);
-        wl_surface_commit(window->surface);
     } else {
         assert(0);
     }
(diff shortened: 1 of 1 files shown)


aleasto (approved):

aleasto:
Mutter isn't happy when launching an app in windowed mode when the session wasn't previously running.
```
(mutter:194071): libmutter-WARNING **: 14:51:25.474: Invalid window geometry for xdg_surface@13. Ignoring for now, but this will result in client termination in the future.
```

dos1:
Hmm, could you log out which values of `width` and `height` are being set? The protocol only says that they have to be greater than zero, so it kinda looks like the window sizing code has been buggy somewhere already and this simply makes it visible.

dos1:
I don't see how this warning could possibly be triggered by this code.

Mutter:
```
      if (priv->geometry.width == 0 || priv->geometry.height == 0)
        {
          g_warning ("Invalid window geometry for xdg_surface@%d. Ignoring "
                     "for now, but this will result in client termination "
                     "in the future.",
                     wl_resource_get_id (priv->resource));
          return;
        }
```

Waydroid:
```
    if (width == 0 || height == 0) {
		/* Compositor is deferring to us */
		return;
	}

    if (! window->display->isWinResSet) {
        xdg_surface_set_window_geometry(window->xdg_surface, 0, 0, width, height);
```

Are you sure this is Waydroid's surface that triggers this?

I have pushed a new version that moved this call into a more appropriate place, but this shouldn't affect what you're seeing at all.

aleasto:
It is definitely waydroid triggering it.
Before ending up in `priv->geometry` it goes through `meta_wayland_shell_surface_determine_geometry` which I suppose makes them zero?

I'm running a nested mutter wayland window:
```bash
MUTTER_DEBUG_DUMMY_MODE_SPECS=1200x800  mutter --nested --wayland
```
Printing the values from `wayland-hwc.cpp` I get back `1200x800`

dos1:
> Before ending up in priv->geometry it goes through meta_wayland_shell_surface_determine_geometry which I suppose makes them zero?

That would likely mean a bug in Mutter, as I can't think of any way a surface can make its effective geometry to be zero (you can't send a buffer with zero size nor set a viewport with zero destination, that would be a hard protocol violation).

dos1:
Oh, maybe Mutter does that because there's no buffer attached at a time when set_window_geometry is called? That would still be a bug in Mutter, as I don't see anything in the protocol that would prevent clients from doing that.

dos1:
Seems like they are reading the protocol in a different way than I do: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/54

dos1:
OK, made a change that should make Mutter happy now. Turned out that the background color stuff made it easier than it would be before :)

aleasto:
Mutter is indeed happy now. Thanks!

aleasto:
@dos1 dc52721b4067b6aafde7a90def363259a3088bed breaks wlroots:
```
10-31 22:01:22.694    68   102 E wayland-hwc: xdg_surface@18: error 3: xdg_surface has never been configured
10-31 22:01:22.694    68   102 E hwcomposer: *** void *hwc_wayland_thread(void *): Wayland client was disconnected: Unknown error -1
```

Found this useful discussion: https://gitlab.freedesktop.org/gstreamer/gst-plugins-bad/-/issues/1290
Where `From IRC, I am told that wl_surface.attach MUST NOT HAPPEN BEFORE commit()`

Thoughts? :/

dos1:
My primary testing environment is a wlroots-based compositor (phoc) - could you add more details on what exactly gets broken?

Quackdoc:
brakes multi window

waydroid prop set persist.waydroid.multi_windows true

aleasto:
> My primary testing environment is a wlroots-based compositor (phoc) - could you add more details on what exactly gets broken?

With `persist.waydroid.use_subsurface` or `persist.waydroid.multi_windows` Wayfire and Sway both break.

PR #20, opened 2022-08-19
hwcomposer: Allow to use subsurfaces with single window mode

dos1:
This creates a new prop "persist.waydroid.use_subsurface" that allows to turn subsurface usage on regardless of whether multi window mode is enabled.

### DIFF
--- a/hwcomposer/hwcomposer.cpp
+++ b/hwcomposer/hwcomposer.cpp
@@ -70,6 +70,7 @@ struct waydroid_hwc_composer_device_1 {
     int timeline_fd;
     int next_sync_point;
     bool use_subsurface;
+    bool multi_windows;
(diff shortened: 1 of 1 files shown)


aleasto (approved):
Not sure what's the purpose but i don't mind merging it

dos1:
Two purposes:
- using subsurfaces offloads the composition into Wayland compositor, which allows it to optimize composition out in some scenarios by using scanout planes just like Android's hwc would do on real hardware, offering potential performance and/or power consumption improvements
- for some yet unknown reason, composition within Android doesn't seem to work on the Librem 5 right now and results only in fully transparent buffers, while subsurface mode works fine, so this was an easy way to make it work quick :)

PR #21, opened 2022-08-19
hwcomposer: Fix explicit sync and use it with subsurfaces too

dos1:
Release fences can't use a shared timeline, because they're unrelated events that can happen out of order.

Also, move retire fence creation at the end of hwc_set, so its descriptor doesn't have to be closed on every early return, and make the retire fence actually be signalled when commiting the next composition on Wayland surface.

### DIFF
--- a/hwcomposer/hwcomposer.cpp
+++ b/hwcomposer/hwcomposer.cpp
@@ -337,7 +337,6 @@ static int hwc_set(struct hwc_composer_device_1* dev,size_t numDisplays,
     }
 
     hwc_display_contents_1_t* contents = displays[HWC_DISPLAY_PRIMARY];
-    contents->retireFenceFd = sw_sync_fence_create(pdev->timeline_fd, "hwc_contents_release", pdev->next_sync_point);
 
     if (pdev->display->geo_changed) {
         for (auto it = pdev->display->buffer_map.begin(); it != pdev->display->buffer_map.end(); it++) {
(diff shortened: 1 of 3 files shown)


aleasto (approved):

aleasto:
So this breaks `persist.waydroid.use_subsurface=true` and `persist.waydroid.multi_windows=true` on ubuntu only. huh?

It causes "Application Not Responding"

aleasto:
So it looks like ubuntu might be the only distro shipping kernels with `CONFIG_SW_SYNC`, that `libsync` uses to implement the APIs used here.

That raises two questions:
Is any of this necessary?
And how did you test it?

This diff unbreaks it, hinting that the `sw_sync_timeline_inc` in `wl_buffer::release` might not be firing correctly
```diff
diff --git a/hwcomposer/hwcomposer.cpp b/hwcomposer/hwcomposer.cpp
index 823b531..3db3923 100644
--- a/hwcomposer/hwcomposer.cpp
+++ b/hwcomposer/hwcomposer.cpp
@@ -772,6 +772,7 @@ static int hwc_set(struct hwc_composer_device_1* dev,size_t numDisplays,
             int timeline_fd = sw_sync_timeline_create();
             /* To be signaled when the compositor releases the buffer */
             fb_layer->releaseFenceFd = sw_sync_fence_create(timeline_fd, "wayland_release", 1);
+            sw_sync_timeline_inc(timeline_fd, 1);
             buf->timeline_fd = timeline_fd;
         } else {
             buf->timeline_fd = -1;
```

aleasto:
~In facts when writing `buf->timeline_fd = timeline_fd;` the previous value was not `-1` as it should be if everything was synchronized correctly.~
EDIT: nevermind, this is because apparently every frame has HWC_GEOMETRY_CHANGED, and with software rendering that implies recreating the `struct buffer`. Also odd.

dos1:
> So it looks like ubuntu might be the only distro shipping kernels with CONFIG_SW_SYNC, that libsync uses to implement the APIs used here.

That would explain so much :laughing: 

The previous sync implementation was clearly broken, so I reworked it as an attempt to fix outdated buffers flickering in subsurface mode, but in the end it turned out to be unrelated and it was #14 that fixed it.

I don't think there's a lot of value in trying to fake explicit sync there. It could be disabled and maybe revisited one day with something like [linux-explicit-synchronization-unstable-v1](https://wayland.app/protocols/linux-explicit-synchronization-unstable-v1) - but not before it gets actually supported across the stack ;)

PR #23, opened 2022-08-28
crop on any gralloc

aleasto:


aleasto:
With surfaceflinger patch:

```diff
diff --git a/services/surfaceflinger/DisplayHardware/ComposerHal.cpp b/services/surfaceflinger/DisplayHardware/ComposerHal.cpp
index 107e34cc5f..afef7188d3 100644
--- a/services/surfaceflinger/DisplayHardware/ComposerHal.cpp
+++ b/services/surfaceflinger/DisplayHardware/ComposerHal.cpp
@@ -607,7 +607,7 @@ Error Composer::setClientTarget(Display display, uint32_t slot,
     const native_handle_t* handle = nullptr;
     if (target.get()) {
         if (mWaydroidDisplay)
-            mWaydroidDisplay->setTargetLayerHandleInfo(target->getPixelFormat(), target->getStride());
+            mWaydroidDisplay->setTargetLayerHandleInfo(target->getPixelFormat(), target->getStride(), target->getWidth(), target->getHeight());
         handle = target->getNativeBuffer()->handle;
     }
 
@@ -743,7 +743,7 @@ Error Composer::setLayerBuffer(Display display, Layer layer,
     if (buffer.get()) {
         if (mWaydroidDisplay)
             mWaydroidDisplay->setLayerHandleInfo(mLayersZMap[layer], buffer->getPixelFormat(),
-                                              buffer->getStride());
+                                              buffer->getStride(), buffer->getWidth(), buffer->getHeight());
         handle = buffer->getNativeBuffer()->handle;
     }
 ```

Since we distribute system and vendor images separately this needs a version bump to IWaydroidDisplay and surfaceflinger should handle both versions... But before spending time on that maybe there's another solution?

@dos1

aleasto:
Also this still doesn't solve a SEGV_MAPPER fault I'm hitting on gralloc.default in subsurface mode at https://github.com/waydroid/android_hardware_waydroid/blob/lineage-18.1/hwcomposer/hwcomposer.cpp#L124

dos1:
Looks reasonable to me; I believe the only way to make it saner would be to use Gralloc/Mapper 4.

> Also this still doesn't solve a SEGV_MAPPER fault I'm hitting on gralloc.default in subsurface mode

Is this something that was introduced by my changes, or was it happening in multiwindow mode earlier as well?

aleasto:
> > Also this still doesn't solve a SEGV_MAPPER fault I'm hitting on gralloc.default in subsurface mode
> 
> Is this something that was introduced by my changes, or was it happening in multiwindow mode earlier as well?

Was happening earlier as well. But I was hoping using the correct dimensions would fix this presumably out of bounds read.

Reproducible by just launching apps or toggling maximize/unmaximize in multi windows mode

aleasto:
Pushed cleaner version

aleasto:
Multi-windows now works on `gralloc.default`!

PR #24, opened 2022-08-28
gralloc: Implement CrOS API to retrieve buffer info

dos1:
This allows Mesa to use correct buffer modifiers.

### DIFF
--- a/gralloc/gralloc.cpp
+++ b/gralloc/gralloc.cpp
@@ -33,11 +33,24 @@
 
 #include <hardware/gralloc.h>
 #include <system/graphics.h>
+#include <android/gralloc_handle.h>
+#include <drm/drm_fourcc.h>
 
 #include <gbm.h>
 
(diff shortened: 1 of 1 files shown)


gralloc/gralloc.cpp
```diff
@@ -64,12 +77,38 @@ static int gbm_init(struct gbm_module_t *dmod)
 	return err;
 }
 
+static int get_fourcc(int format)
```
aleasto, line 80: Should this account for `persist.waydroid.invert_colors` like `get_gbm_format`?
Also gbm fourcc codes seem to be identical to drm, so we could directly use `get_gbm_format`

dos1: Isn't the point of `persist.waydroid.invert_colors` to lie about used format?

aleasto: i don't know :shrug:

PR #31, opened 2022-08-29
hwcomposer: Simulate a key press on window close

dos1:
This is meant to restart the idle timer, so suspending can work even after closing the window with idle timeout already passed.
See https://github.com/waydroid/waydroid/issues/500

Initially I wanted to use pointer movement, but that doesn't work when there's no pointer (which, well, is not unusual on a phone).

Is there a better way to restart idle timers? Maybe something like taking screen wakelock and releasing it immediately? I don't know much about Android, so for now I have simply used something that was already at hand ;)

### DIFF
--- a/hwcomposer/wayland-hwc.cpp
+++ b/hwcomposer/wayland-hwc.cpp
@@ -313,9 +313,16 @@ xdg_toplevel_handle_configure(void *data, struct xdg_toplevel *,
 }
 
 static void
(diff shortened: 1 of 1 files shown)


dos1:
@aleasto Any ideas how to deal with it in a better way? I have hacked the Python side to shut the container down on suspend to save resources instead of just freezing it and having this working reliably would be so much better than having to remember to shut the container down manually.

aleasto:
I wouldn't know how else to do it so I'm ok with this.

PR #33, opened 2022-08-30
hwcomposer: Don't try to use viewporter when not available

### LINKED ISSUE #32: android.hardware.graphics.composer@2.1-service crashes on Ubuntu Touch
Adding as a tracking issue so this isn't forgotten about, @dos1 you've been doing a lot of `hwcomposer` related work recently so you should know about this I believe. What were those patches tested on?

I tested this on a Volla Phone 22 / Halium 11 running Ubuntu Touch 16.04 (2022-08-22 / devel channel) with Waydroid 1.2.0 and Android 11 images (so using [`lineage-18.1` branch](https://github.com/waydroid/android_hardware_waydroid/commits/lineage-18.1) here) provided by @aleasto.

He suspected [Mir](https://github.com/MirServer/mir) (2.9.0 even) is missing some required Wayland protocol support, specifically https://wayland.app/protocols/viewporter, so my question is: should I be able to see a similar crash on x86_64 when running the Mir 2.9.0 test compositor as seen in the `logcat` from the phone attached below?
[waydroid-logcat.log](https://github.com/waydroid/android_hardware_waydroid/files/9448430/waydroid-logcat.log)
[waydroid-log.log](https://github.com/waydroid/android_hardware_waydroid/files/9448429/waydroid-log.log)

Fwiw Ubuntu Touch uses a [patched Mir 1.8.2 version](https://gitlab.com/ubports/development/core/packaging/mir/-/tree/ubports/focal/debian/patches) and testing that outside of a device running Ubuntu Touch might be tricky or not possible (I've not tried yet).

Related(?): #12

dos1:
Closes #32

### DIFF
--- a/hwcomposer/hwcomposer.cpp
+++ b/hwcomposer/hwcomposer.cpp
@@ -231,7 +231,8 @@ static struct wl_surface *get_surface(struct waydroid_hwc_composer_device_1 *pde
         subsurface = wl_subcompositor_get_subsurface(pdev->display->subcompositor,
                                                      surface,
                                                      window->surface);
-        viewport = wp_viewporter_get_viewport(pdev->display->viewporter, surface);
+        if (pdev->display->viewporter)
+            viewport = wp_viewporter_get_viewport(pdev->display->viewporter, surface);
         window->surfaces[window->lastLayer] = surface;
         window->subsurfaces[window->lastLayer] = subsurface;
         window->viewports[window->lastLayer] = viewport;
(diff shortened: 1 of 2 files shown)


aleasto:
Huh that's all?

dos1:
I think so... What else do you expect? :smile:

aleasto:
> Of course crops will stay broken on such compositors, but at least the background can still be provided by creating a full-sized buffer instead.

That?

dos1:
That's something that can be done later. This background wasn't there previously at all so it's not like something's getting broken and, personally, I only really care about how it works under phoc - so I have fixed my code to not break other use-cases than mine, but actually improving those other use-cases is probably a task for someone else (unless I'll get bored, I guess) :wink:

PR #40, opened 2022-10-17
hwcomposer: refactor hwc_set into smaller functions

DakeuTaepi:
set() is a HWComposer HAL method which gets called when SurfaceFlinger is ready to present its layers on screen (https://android.googlesource.com/platform/hardware/libhardware/+/master/include/hardware/hwcomposer.h#531). Waydroid seems to use it as main place to hijack rendering and map SurfaceFlinger layers (with extra metadata like task ID/app ID added) to Wayland windows.

Unfortunately, because there at least 3 different windowing modes supported (render Android screen in a single window, create a distinct Android screen window per app, and desktop free-form multi-window mode), the hwc_set functions looks pretty convoluted and hard to extend/modify.

I've attempted to slim it down a bit by separating code chunks to make hwc_set() easier to read, but not quite happy with the result yet. Maybe someone else would have a better idea how to split it up after reading the code, so any feedback would be welcome.

### DIFF
--- a/hwcomposer/hwcomposer.cpp
+++ b/hwcomposer/hwcomposer.cpp
@@ -75,6 +75,8 @@ struct waydroid_hwc_composer_device_1 {
     bool multi_windows;
 };
 
+const std::string waydroid_window_task_id = "0";
+
 static int hwc_prepare(hwc_composer_device_1_t* dev,
                        size_t numDisplays, hwc_display_contents_1_t** displays) {
     struct waydroid_hwc_composer_device_1 *pdev = (struct waydroid_hwc_composer_device_1 *)dev;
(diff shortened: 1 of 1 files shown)


hwcomposer/hwcomposer.cpp
```diff
@@ -383,7 +385,224 @@ static const struct wp_presentation_feedback_listener feedback_listener = {
...
+        auto const& window_task_id = it->first;
+        auto& window = it->second;
+
+        bool should_erase = false;
+
+        if (active_layer_task_id == waydroid_window_task_id) {
+            // Screen in a window mode
+            if (window_task_id != waydroid_window_task_id) {
+                // Close everything except for the screen window
+                should_erase = true;
```
maciek134: `should_erase = (window_task_id != waydroid_window_task_id);` would reduce nesting here

hwcomposer/hwcomposer.cpp
```diff
@@ -383,7 +385,224 @@ static const struct wp_presentation_feedback_listener feedback_listener = {
...
+            }
+        } else {
+            // Single window per task mode
+            if (window_task_id == waydroid_window_task_id
+                || (active_layer_task_id != window_task_id && !window->isActive)) {
+                // Close previously shown whole screen window and remove
+                // non-current layer window that has been closed by compositor
+                should_erase = true;
```
maciek134: And 
```
should_erase = (window_task_id == waydroid_window_task_id
                || (active_layer_task_id != window_task_id && !window->isActive));
```

aleasto (changes requested):
Thank you for the contribution.

Regressions found on my end (kwin_wayland 5.26.2):

- Background in multi-windows mode is not transparent anymore
- The `persist.waydroid.cursor_on_subsurface` hack for showing the cursor on kwin_wayland in multi-windows mode doesn't work anymore

PR #42, opened 2022-11-01
Xdg surface fixes

aleasto:


aleasto:
Makes both wlroots (`xdg_surface@18: error 3: xdg_surface has never been configured`) and mutter (`Invalid window geometry for xdg_surface@13. Ignoring for now, but this will result in client termination in the future.`) happy

PR #43, opened 2022-11-10
hwcomposer: Correctly handle custom display size from the user

aleasto:
If the user has requested a certain display size, make the window not
maximized and trim the background and all regions to the specified size

If the user has requested padding instead, keep the window maximized
and keep drawing the background on the pad area

### DIFF
--- a/hwcomposer/wayland-hwc.cpp
+++ b/hwcomposer/wayland-hwc.cpp
@@ -287,6 +287,30 @@ static const struct xdg_surface_listener xdg_surface_listener = {
     xdg_surface_handle_configure,
 };
 
+static void choose_width_height(struct display* display, int32_t hint_width, int32_t hint_height) {
+    char property[PROPERTY_VALUE_MAX];
+    int width = hint_width;
+    int height = hint_height;
+
+    // Ignore hint it requested
+    if (property_get("persist.waydroid.width", property, nullptr) > 0) {
+        display->isMaximized = false;
+        width = atoi(property);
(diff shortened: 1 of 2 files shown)


aleasto:
This is a regression after https://github.com/waydroid/android_hardware_waydroid/pull/12, https://github.com/waydroid/android_hardware_waydroid/pull/18

aleasto:
@dos1 does this look right?

dos1:
xdg-shell says this about `maximized` state:

> The surface is maximized. The window geometry specified in the configure event must be obeyed by the client.

You can't set `xdg_toplevel_set_maximized` if you want to size the window on your own.

aleasto:
Thanks.
I don't think I need to do anything else because the actual window size will be set by `xdg_surface_set_window_geometry`, when using `xdg-shell` at least.

I did try to run our `wl_shell` backend instead but that seems to be broken in the current git HEAD already

aleasto:
And of course, now that I wrote that I realize that we're calling `xdg_surface_set_window_geometry` even on `wl_shell` and that's probably why it's broken

aleasto:
I can't seem to find an alternative to xdg_surface_set_window_geometry for wl_shell, but it seems to be doing ok without.

aleasto:
Mh, how do I ensure the window starts at 0,0 if it's not maximized?

aleasto:
I don't think that's left up to clients in wayland right...? Need to rely on the compositor to make a good decision

PR #45, opened 2022-12-23
health_service: Disable periodic chores

### LINKED ISSUE #168: [Pinephone] Waydroid interferes with suspend/deep sleep
I wasn't sure if it was just my device that experienced this problem but some others on Reddit have run into this as well, e.g.: https://www.reddit.com/r/pinephone/comments/qb3vo3/comment/hh77s7p/?context=3

On the Pinephone, the Crust firmware can suspend the phone to deep sleep to conserve battery life.

If the Waydroid session is running, it seems to interfere with deep sleep:

* My phone's suspend TTL is 60 seconds and the phone will go to sleep as normal.
* After several minutes (5 or 10) the screen will turn back on and show my lock screen.
* I can press the power button to turn the screen back off and the phone will nap, but another 5-10 minutes later the screen will turn back on.
* If I don't do anything to it, the screen stays on forever and would eat thru its battery.

If I fully reboot the phone and then _do not_ launch any Android app, suspend works as expected, the phone stays asleep unless the cell modem awakens it.

I saw this on Mobian with Phosh and others have seen it on Manjaro and Arch as well.

dos1:
Fake battery status doesn't change, so there's nothing for healthd   to check periodically. Since healthd is using CLOCK_BOOTTIME_ALARM timers, this was causing the host to wake up from suspend very  often (the default timeout in effect was 60 seconds).

Fixes https://github.com/waydroid/waydroid/issues/168

**NOTE**: I haven't tested this. If you want to save me from having to spin up another Android image build, feel free to test it - I'll be grateful:)

### DIFF
--- a/health/health_service.cpp
+++ b/health/health_service.cpp
@@ -24,7 +24,10 @@
 using android::hardware::health::V2_0::DiskStats;
 using android::hardware::health::V2_0::StorageInfo;
 
-void healthd_board_init(struct healthd_config*) {}
+void healthd_board_init(struct healthd_config* config) {
+  config->periodic_chores_interval_fast = -1;
+  config->periodic_chores_interval_slow = -1;
+}
 
 int healthd_board_battery_update(
     struct android::BatteryProperties* battery_props) {


health/health_service.cpp
```diff
@@ -24,7 +24,10 @@
 using android::hardware::health::V2_0::DiskStats;
 using android::hardware::health::V2_0::StorageInfo;
 
-void healthd_board_init(struct healthd_config*) {}
+void healthd_board_init(struct healthd_config*) {
```
aleasto:
```suggestion
void healthd_board_init(struct healthd_config* healthd_config) {
```

dos1: Ha :D

aleasto:
I'm not sure this is it for issue https://github.com/waydroid/waydroid/issues/168
I think I can reproduce on my laptop, and this commit doesn't help

dos1:
Could be that it's not sufficient and there are still more timers to disable somewhere - but I'm pretty sure this will be necessary, since it's using `CLOCK_BOOTTIME_ALARM` timer directly: https://cs.android.com/android/platform/superproject/+/master:hardware/interfaces/health/utils/libhealthloop/HealthLoop.cpp?q=CLOCK_BOOTTIME_ALARM&start=91

Seems like the nuclear option would be to patch AlarmManagerService to not use _ALARM timers: https://cs.android.com/android/platform/superproject/+/master:frameworks/base/apex/jobscheduler/service/jni/com_android_server_alarm_AlarmManagerService.cpp?q=CLOCK_BOOTTIME_ALARM&start=101 - alternatively, we'd have to find all uses of RTC_WAKEUP and REALTIME_WAKEUP alarms and see which ones should be disabled.

PR #47, opened 2023-01-05
hwcomposer: Snapshot inactive windows surface (v2)

aleasto:


aleasto:
~I've seen this randomly fail but am unable to reproduce.~

~EDIT: Calling `egl_init` every time makes it stable. I'm not sure what I'm doing wrong.~

Done. hwc_set might be called from a different binder thread while gl context is thread local

PR #48, opened 2023-01-29
hwcomposer: Synchronization fixes

aleasto:


aleasto:
@dos1

This almost works. But sometimes it breaks when switching from full-ui mode to single-app mode

aleasto:
Well, that is caused by contents->retireFenceFd, not the layers one.

hwcomposer/hwcomposer.cpp
```diff
@@ -191,8 +191,8 @@ static struct buffer *get_wl_buffer(struct waydroid_hwc_composer_device_1 *pdev,
         if (it->second->isShm) {
             if (width != it->second->width || height != it->second->height) {
                 if (it->second->buffer)
-                    wl_buffer_destroy(it->second->buffer);
-                delete (it->second);
+                    it->second->destroy = true;
+                else delete (it->second);
```
aleasto: This leaks if we reach this point after wl_buffer_release

Man this is hard

aleasto:
So, I believe this logic is correct.

**However** it relies on the assumption that the compositor will always reliably send a `wl_buffer::release` event, even when replacing a surface's buffer with itself again. Mutter curiously seems to respect this, but i think it is not the ideal behaviour; indeed kwin and weston do it differently.

So I believe I'm just going to keep this here to be revisited, and for now disable the per-layer synchronization.

aleasto:
Alternatively, since the buffer cache is very rarely used as most frames post-boot have the `GEOMETRY_CHANGED` flag, we could just remove it and then `wl_buffer::release` should behave how this logic requires.

PR #49, opened 2023-02-10
hwcomposer: Implement touch_handle_cancel

roscale:
If the compositor decided to send a touch cancel event instead of an up event, the pointer would permanently stay down inside Waydroid, and restarting the container was the only way to fix it.

Now, pointers are properly canceled by turning them into palms before lift-off. I don't know how reasonable is this solution, but it works.

To give you some context, I'm writing my own compositor and I have gesture navigation similar to Android. Swiping up from the bottom reveals the task switcher. I need touch cancel to work because initially, the application captures the touch down event, but if a swipe up is detected, it sends a cancel event to the application. The compositor then takes over and captures the remaining touch events for that pointer to reveal the task switcher.

### DIFF
--- a/hwcomposer/wayland-hwc.cpp
+++ b/hwcomposer/wayland-hwc.cpp
@@ -1069,8 +1069,44 @@ touch_handle_frame(void *, struct wl_touch *)
 }
 
 static void
-touch_handle_cancel(void *, struct wl_touch *)
+touch_handle_cancel(void *data, struct wl_touch *)
 {
+    struct display* display = (struct display*)data;
+    struct input_event event[6];
+    struct timespec rt;
+    unsigned int res, n;
+    int i, id;
+
+    if (ensure_pipe(display, INPUT_TOUCH))
+        return;
+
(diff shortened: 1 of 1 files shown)


aleasto (approved):
How may I test this on like mutter or kwin etc...?
Regardless, LGTM

roscale:
I'll try to see how I can trigger this bug under kwin or mutter and I will write the steps to reproduce.

roscale:
Steps to reproduce:
- Run Waydroid on Gnome Shell.
- Swipe up using 3 fingers to open the activities overview.

The first 2 finger down events are captured by Waydroid, but when the 3rd finger touches the screen, Mutter sends a touch cancel event. Waydroid still thinks the first 2 fingers are down.

To see more easily what I'm talking about, using the Android browser go to a web page that performs multitouch tests, like this one https://tecagile.com/touch-results/ and try the gesture here.

Then compare the behavior using this commit.

aleasto:
Thanks

PR #52, opened 2023-09-04
Support explicit dmabuf modifiers from the wayland compositor

aleasto:


aleasto:
This doesn't work yet.

Also, we need to treat system properties more carefully (length limit, cache them locally)

aleasto:
I was hoping this would resolve the artifacts we see on multi-gpu systems when waydroid renders on one gpu and the wayland compositor on another one, but apparently not -- or not in every case.

@technobaboo Are you able to test whether at least this addresses your issue, if i provide you with a build that includes this diff?

technobaboo:
yep! is that an image I can replace my current one with?

aleasto:
[x86_64.tar.gz](https://github.com/waydroid/android_hardware_waydroid/files/12611315/x86_64.tar.gz)

We use overlayfs (if your kernel supports it) so `sudo mount -o rw,remount /var/lib/waydroid/rootfs` and replace the files in `/var/lib/waydroid/rootfs/vendor/lib64/hw/` then restart Waydroid

The changes will be undone the next OTA update of the vendor image, or manually by removing `/var/lib/waydroid/overlay_rw/`

aleasto:
By `sudo waydroid shell getprop | grep modifiers` you should find the list of modifiers per gbm format (hex encoded) that gralloc will use after filtering out those that would result in a multi-planar buffer which we don't support.

Mind that we implement version 3 of the zwp_linux_dmabuf_v1 protocol and not version 4

technobaboo:
```
❯ sudo mount -o rw,remount /var/lib/waydroid/rootfs
[sudo] password for nova:
mount: /var/lib/waydroid/rootfs: mount point not mounted or bad option.
       dmesg(1) may have more information after failed mount system call.
```

sorry, i'm very new to this (this is with the waydroid session started and system service running)

aleasto:
Doesn't make much sense, if waydroid is running.

Please reach me at `aleasto` on telegram, matrix.org, libera.chat or oftc.net

aleasto:
TODO: Implement this in gralloc.minigbm_gbm_mesa too

PR #55, opened 2024-03-25
Fix for issue playing YouTube videos through the built-in Chromium browser

eagleivg:
The problem is described in more detail here: https://github.com/waydroid/waydroid/discussions/855. 

It was found that when using pulse PCM, snd_pcm_htimestamp returns correct data, judging by the description:

out_get_presentation_position: ret = 0 , frames = 1165312, ts=1711351920.591045044

But at the same time the playback of the video stops on the initial frame, and the sound is looped on the playback of the first 10 seconds. In other players there are no problems, personally I have Fennec on arm64 played video correctly, although in the discussion other users complained about errors on x64_86, players like NewPipe and VLC had no problems.

If I changed PCM to hardware PCM (having previously thrown it into the container), the video started playing correctly with sound. However, snd_pcm_htimestamp returned zero timestamp in this case:

audio_hw_primary: out_get_presentation_position: ret = 0 , frames = 1478708, ts=0.000000000

Also, replacing audio.primary.waydroid.so with the default null library returned correct playback
Therefore, the simplest solution is to remove the out_get_presentation_position callback.

As far as I could check, this solution has no side effects; all the programs I tested played video and sound correctly. And this is easier than finding the reason for the incorrect operation of chromium or searching for the peculiarities of how snd_pcm_htimestamp works in a container and without it.

### DIFF
--- a/audio/audio_hw.c
+++ b/audio/audio_hw.c
@@ -502,28 +502,6 @@ static int out_get_render_position(const struct audio_stream_out *stream,
     return -EINVAL;
 }
 
-static int out_get_presentation_position(const struct audio_stream_out *stream,
-                                   uint64_t *frames, struct timespec *timestamp)
-{
-    struct alsa_stream_out *out = (struct alsa_stream_out *)stream;
-    int ret = -ENODATA;
-
-        if (out->pcm) {
-            snd_pcm_uframes_t avail;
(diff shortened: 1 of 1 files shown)


aleasto:
Our libalsa/pulse are not exactly up-to-date; it might be that updating those libraries would fix this.

In the meantime we can definitely give this a try. Thanks!

PR #58, opened 2024-09-12
hwcomposer: wp_cursor_shape_manager_v1 implements

atmouse-:
protocol [android_vendor_waydroid/pull/37](https://github.com/waydroid/android_vendor_waydroid/pull/37) requires

This fix invisible cursor if that window focus switch from playing fullscreen video out of waydroid


aleasto:
Cursor handling has changed a lot since 2024. And this never made too much sense anyways.

PR #59, opened 2025-01-14
[AudioHAL] Pass frames instead of bytes to ALSA `set_buffer_size()`

sabae-valve:
According to the ALSA documentation [0], this method takes frames in as the third parameter, however our `out_get_buffer_size()` function returns bytes.  We do the conversion here to ensure that both sides agree on the relevant buffer sizes.

Note that in our experiments, we were able to achieve significantly lower latency with this patch and by reducing `PLAYBACK_PERIOD_SIZE` to `512`, however we recognize this may not be desirable on all hardware, and so have omitted that change here.

[0] https://www.alsa-project.org/alsa-doc/alsa-lib/group___p_c_m___h_w___params.html#ga828cf9e11fe84137c2c8d4938a6fca7f

### DIFF
--- a/audio/audio_hw.c
+++ b/audio/audio_hw.c
@@ -224,8 +224,10 @@ static int start_output_stream(struct alsa_stream_out *out)
   	    return -ENODEV;
     }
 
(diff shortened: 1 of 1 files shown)


aleasto:
Thank you! Works well.
I've fixed up the printf format specified for `size_t`.

I'm also going to try reducing the period_size to 1024 at least.

PR #63, opened 2025-05-27
Rework/Cleanup hwcomposer

Molytho:
Here is my first take on hwcomposer cleanup.
I'm not finished but thought I should commit the first bunch since I already got 72 commits.
Most commits are small and self-explaining but there are also some rather big ones.

I'm currently not aware of any regressions but I only have a sway setup and did not test my changes on gnome and kde. Input is welcome.

### DIFF
--- a/hwcomposer/gralloc_handler.cpp
+++ b/hwcomposer/gralloc_handler.cpp
@@ -114,11 +114,7 @@ namespace {
     };
 
     bool isFormatSupported(struct display *display, uint32_t format) {
-        for (int i = 0; i < display->formats_count; i++) {
-            if (format == display->formats[i])
-                return true;
-        }
-        return false;
+        return display->formats.count(format) == 1;
     }
 
     uint32_t ConvertHalFormatToDrm(struct display *display, uint32_t hal_format) {
--- a/hwcomposer/gralloc_handler.cpp
+++ b/hwcomposer/gralloc_handler.cpp
@@ -74,12 +74,14 @@ std::unique_ptr<buffer> create_shm_wl_buffer(display *display, const buffer_meta
     int shm_stride = metadata.width * 4;
     int size = shm_stride * metadata.height;
 
-    buf->size = size;
     buf->metadata = metadata;
-    buf->format = ConvertHalFormatToShm(metadata.format);
-    assert(buf->format >= 0);
     buf->handle = handle;
+
     buf->isShm = true;
+    buf->size = size;
+
+    auto shm_format = ConvertHalFormatToShm(metadata.format);
+    assert(shm_format >= 0);
 
     int fd = syscall(SYS_memfd_create, "buffer", MFD_ALLOW_SEALING);
     ftruncate(fd, size);
@@ -90,7 +92,7 @@ std::unique_ptr<buffer> create_shm_wl_buffer(display *display, const buffer_meta
         return nullptr;
     }
     struct wl_shm_pool *pool = wl_shm_create_pool(display->shm, fd, size);
-    buf->wl_buffer = wl_shm_pool_create_buffer(pool, 0, metadata.width, metadata.height, shm_stride, buf->format);
+    buf->wl_buffer = wl_shm_pool_create_buffer(pool, 0, metadata.width, metadata.height, shm_stride, shm_format);
     wl_buffer_add_listener(buf->wl_buffer, &buffer_listener, nullptr);
     wl_shm_pool_destroy(pool);
     close(fd);
@@ -173,15 +175,18 @@ std::unique_ptr<buffer> create_dmabuf_wl_buffer(display *display, const buffer_m
     std::unique_ptr<buffer> buf { new buffer() };
 
     buf->metadata = metadata;
-    buf->format = (drm_format >= 0) ? drm_format : ConvertHalFormatToDrm(display, metadata.format);
-    assert(buf->format >= 0);
     buf->handle = handle;
+
+    if (drm_format < 0) {
+        drm_format = ConvertHalFormatToDrm(display, metadata.format);
+    }
+    assert(drm_format >= 0);
 
     zwp_linux_buffer_params_v1 *params = zwp_linux_dmabuf_v1_create_params(display->dmabuf);
     zwp_linux_buffer_params_v1_add(params, prime_fd, 0, offset, byte_stride, modifier >> 32, modifier & 0xffffffff);
     zwp_linux_buffer_params_v1_add_listener(params, &params_listener, nullptr);
 
-    buf->wl_buffer = zwp_linux_buffer_params_v1_create_immed(params, buf->metadata.width, buf->metadata.height, buf->format, 0);
+    buf->wl_buffer = zwp_linux_buffer_params_v1_create_immed(params, buf->metadata.width, buf->metadata.height, drm_format, 0);
     wl_buffer_add_listener(buf->wl_buffer, &buffer_listener, nullptr);
 
     return buf;
@@ -192,7 +197,6 @@ std::unique_ptr<buffer> create_android_wl_buffer(display *display, const buffer_
     std::unique_ptr<buffer> buf { new buffer() };
 
     buf->metadata = metadata;
-    buf->format = metadata.format;
     buf->handle = handle;
 
     struct wl_array ints;
@@ -206,7 +210,7 @@ std::unique_ptr<buffer> create_android_wl_buffer(display *display, const buffer_
         android_wlegl_handle_add_fd(wlegl_handle, handle->data[i]);
     }
 
-    buf->wl_buffer = android_wlegl_create_buffer(display->android_wlegl, buf->metadata.width, buf->metadata.height, buf->metadata.pixel_stride, buf->format, GRALLOC_USAGE_HW_RENDER, wlegl_handle);
+    buf->wl_buffer = android_wlegl_create_buffer(display->android_wlegl, buf->metadata.width, buf->metadata.height, buf->metadata.pixel_stride, metadata.format, GRALLOC_USAGE_HW_RENDER, wlegl_handle);
     android_wlegl_handle_destroy(wlegl_handle);
 
     wl_buffer_add_listener(buf->wl_buffer, &buffer_listener, nullptr);
--- a/hwcomposer/gralloc_handler.cpp
+++ b/hwcomposer/gralloc_handler.cpp
@@ -1,0 +1,347 @@
+/*
+ * Copyright © 2025 Waydroid Project.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files (the
+ * "Software"), to deal in the Software without restriction, including
+ * without limitation the rights to use, copy, modify, merge, publish,
+ * distribute, sublicense, and/or sell copies of the Software, and to
+ * permit persons to whom the Software is furnished to do so, subject to
+ * the following conditions:
+ *
+ * The above copyright notice and this permission notice (including the
+ * next paragraph) shall be included in all copies or substantial
+ * portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT.  IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+ * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#include "gralloc_handler.h"
+
+#include <cstdint>
+#include <memory>
+
+#include <linux/memfd.h>
+#include <sys/syscall.h>
+#include <drm_fourcc.h>
+#include <wayland-client.h>
+
+#include <log/log.h>
+#include <ui/Rect.h>
+#include <ui/GraphicBufferMapper.h>
+#include <gralloc_handle.h>
+#include <cros_gralloc/cros_gralloc_handle.h>
+
+#include "linux-dmabuf-unstable-v1-client-protocol.h"
+#include "wayland-android-client-protocol.h"
+
+#include "egl-tools.h"
+#include "wayland-hwc.h"
+
+namespace {
+    uint32_t ConvertHalFormatToShm(uint32_t hal_format) {
+        uint32_t fmt;
+
+        switch (hal_format) {
+            case HAL_PIXEL_FORMAT_RGBX_8888:
+                fmt = WL_SHM_FORMAT_XRGB8888;
+                break;
+            case HAL_PIXEL_FORMAT_RGBA_8888:
+                fmt = WL_SHM_FORMAT_ARGB8888;
+                break;
+            default:
+                ALOGE("Cannot convert hal format to shm format %u", hal_format);
+                return -EINVAL;
+        }
+        return fmt;
+    }
+
+    const wl_buffer_listener buffer_listener {
+        [](void *, struct wl_buffer *) {}
+    };
+}
+std::unique_ptr<buffer> create_shm_wl_buffer(display *display, const buffer_metadata& metadata, buffer_handle_t handle) {
+    std::unique_ptr<buffer> buf { new buffer() };
+
+    // Assume 4bpp formats or none of this is going to work
+    int shm_stride = metadata.width * 4;
+    int size = shm_stride * metadata.height;
+
+    buf->size = size;
+    buf->hal_format = metadata.format;
+    buf->format = ConvertHalFormatToShm(metadata.format);
+    assert(buf->format >= 0);
+    buf->width = metadata.width;
+    buf->height = metadata.height;
+    buf->pixel_stride = metadata.pixel_stride;
+    buf->handle = handle;
+    buf->isShm = true;
+
+    int fd = syscall(SYS_memfd_create, "buffer", MFD_ALLOW_SEALING);
+    ftruncate(fd, size);
+    buf->shm_data = mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0);
+    if (buf->shm_data == MAP_FAILED) {
+        ALOGE("mmap failed");
+        close(fd);
+        return nullptr;
+    }
+    struct wl_shm_pool *pool = wl_shm_create_pool(display->shm, fd, size);
+    buf->buffer = wl_shm_pool_create_buffer(pool, 0, metadata.width, metadata.height, shm_stride, buf->format);
+    wl_buffer_add_listener(buf->buffer, &buffer_listener, nullptr);
+    wl_shm_pool_destroy(pool);
+    close(fd);
+
+    return buf;
+}
+
+namespace {
+    //TODO: Implement failure handling for dmabuf
+    void create_failed(void *data __unused, zwp_linux_buffer_params_v1 *params __unused)
+    {
+        ALOGE("%s: zwp_linux_buffer_params.create_immed failed.", __func__);
+        abort();
+    }
+
+    const struct zwp_linux_buffer_params_v1_listener params_listener = {
+            [](void *, zwp_linux_buffer_params_v1 *, wl_buffer *) {},
+            create_failed
+    };
+
+    bool isFormatSupported(struct display *display, uint32_t format) {
+        for (int i = 0; i < display->formats_count; i++) {
+            if (format == display->formats[i])
+                return true;
+        }
+        return false;
+    }
+
+    uint32_t ConvertHalFormatToDrm(struct display *display, uint32_t hal_format) {
+        uint32_t fmt;
+
+        switch (hal_format) {
+            case HAL_PIXEL_FORMAT_RGB_888:
+                fmt = DRM_FORMAT_BGR888;
+                if (!isFormatSupported(display, fmt))
+                    fmt = DRM_FORMAT_RGB888;
+                break;
+            case HAL_PIXEL_FORMAT_BGRA_8888:
+                fmt = DRM_FORMAT_ARGB8888;
+                if (!isFormatSupported(display, fmt))
+                    fmt = DRM_FORMAT_ABGR8888;
+                break;
+            case HAL_PIXEL_FORMAT_RGBX_8888:
+                fmt = DRM_FORMAT_XBGR8888;
+                if (!isFormatSupported(display, fmt))
+                    fmt = DRM_FORMAT_XRGB8888;
+                break;
+            case HAL_PIXEL_FORMAT_RGBA_8888:
+                fmt = DRM_FORMAT_ABGR8888;
+                if (!isFormatSupported(display, fmt))
+                    fmt = DRM_FORMAT_ARGB8888;
+                break;
+            case HAL_PIXEL_FORMAT_RGB_565:
+                fmt = DRM_FORMAT_BGR565;
+                if (!isFormatSupported(display, fmt))
+                    fmt = DRM_FORMAT_RGB565;
+                break;
+            case HAL_PIXEL_FORMAT_YV12:
+                fmt = DRM_FORMAT_YVU420;
+                if (!isFormatSupported(display, fmt))
+                    fmt = DRM_FORMAT_GR88;
+                break;
+            default:
+                ALOGE("Cannot convert hal format to drm format %u", hal_format);
+                return -EINVAL;
+        }
+        if (!isFormatSupported(display, fmt)) {
+            ALOGE("Current wayland display doesn't support hal format %u", hal_format);
+            return -EINVAL;
+        }
+        return fmt;
+    }
+}
+std::unique_ptr<buffer> create_dmabuf_wl_buffer(display *display, const buffer_metadata& metadata,
+                                                                int prime_fd, int drm_format, int byte_stride,
+                                                                int offset, uint64_t modifier, buffer_handle_t handle)
+{
+    assert(prime_fd >= 0);
+
+    std::unique_ptr<buffer> buf {new buffer()};
+
+    buf->hal_format = metadata.format;
+    buf->format = (drm_format >= 0) ? drm_format : ConvertHalFormatToDrm(display, metadata.format);
+    assert(buf->format >= 0);
+    buf->width = metadata.width;
+    buf->height = metadata.height;
+    buf->pixel_stride = metadata.pixel_stride;
+    buf->handle = handle;
+
+    zwp_linux_buffer_params_v1 *params = zwp_linux_dmabuf_v1_create_params(display->dmabuf);
+    zwp_linux_buffer_params_v1_add(params, prime_fd, 0, offset, byte_stride, modifier >> 32, modifier & 0xffffffff);
+    zwp_linux_buffer_params_v1_add_listener(params, &params_listener, nullptr);
+
+    buf->buffer = zwp_linux_buffer_params_v1_create_immed(params, buf->width, buf->height, buf->format, 0);
+    wl_buffer_add_listener(buf->buffer, &buffer_listener, nullptr);
+
+    return buf;
+}
+
+std::unique_ptr<buffer> create_android_wl_buffer(display *display, const buffer_metadata& metadata, buffer_handle_t handle)
+{
+    std::unique_ptr<buffer> buf {new buffer()};
+
+    buf->width = metadata.width;
+    buf->height = metadata.height;
+    buf->format = buf->hal_format = metadata.format;
+    buf->pixel_stride = metadata.pixel_stride;
+    buf->handle = handle;
+
+    struct wl_array ints;
+    wl_array_init(&ints);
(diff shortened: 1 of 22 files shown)


aleasto:
This is awesome, thank you!

I could only go through ~15 commits today but that is looking very promising; you clearly know what you're doing :)

I'll ask you to please avoid force pushing for some time while I go through all the individual commits, to make my life easier :P

hwcomposer/hwcomposer.cpp
```diff
@@ -470,33 +519,21 @@ static int hwc_set(struct hwc_composer_device_1* dev,size_t numDisplays,
...
                 }
+                return true;
             }
-        }
+            return false;
+        });
+
         // Nothing to show on screen, so clear all open windows
-        if (showWindow == ShowWindowState::BLACKLISTED) {
+        if (!has_unblacklisted_app) {
```
aleasto: I believe this is changing the behaviour.

See commit c16d988:

```
hwcomposer: Only close all windows if the launcher is visible

In single-layer mode, when moving focus from com.android.settings to some other
task, we receive 4 frames without any task layer A13's surfaceflinger.

To avoid losing track of the open windows in this transitioning period,
delay closing windows to when the launcher would be visible.
```

So rather than closing all apps when we didn't match any layer, we'd like to close apps when we **only** matched backlisted layers (meaning that only the home launcher was visible)

To see this bug you would need to use a system image compiled from the lineage-20 branch (you could even use a [prebuilt](https://sourceforge.net/projects/waydroid/files/images/system/lineage/waydroid_x86_64/) system.img, and pair that with a self-compiled lineage-18.1 vendor which does work thanks to project treble)

hwcomposer/hwcomposer.cpp
```diff
@@ -319,6 +319,21 @@ static void apply_surface_damage(hwc_layer_1 *hwc_layer, surface_context &surfac
...
+    const auto& visible_regions = hwc_layer->visibleRegionScreen;
+    auto region = wl_compositor_create_region(pdev->display->compositor);
+    std::for_each(visible_regions.rects, visible_regions.rects + visible_regions.numRects, [&](const auto& rect){
+        int32_t x = static_cast<int32_t>(rect.left / scale);
+        int32_t y = static_cast<int32_t>(rect.top / scale);
+        int32_t width = static_cast<int32_t>(ceil((rect.right - rect.left) / scale));
+        int32_t height = static_cast<int32_t>(ceil((rect.bottom - rect.top) / scale));
+        wl_region_add(region, x, y, width, height);
+    });
+    wl_surface_set_opaque_region(surface_context.surface, region);
```
aleasto: I'm worried that visibleRegionScreen doesn't really map to opaque regions. What about translucent shadows in multi-windows mode?

hwcomposer/wayland-hwc.cpp
```diff
@@ -772,12 +738,11 @@ keyboard_handle_enter(void *data, struct wl_keyboard *,
     display->keyboard_enter_serial = serial;
 
     std::scoped_lock lock(display->windowsMutex);
-    if (display->windows.find(surface) == display->windows.end())
-        return;
-
-    struct window *window = display->windows[surface];
+    auto window = reinterpret_cast<struct window *>(wl_surface_get_user_data(surface));
```
aleasto, line 749: This sounds like a KWin bug, but I'm getting a NULL surface which makes the `wl_surface_get_user_data call` segfault.

Here's how to reproduce:
* Disable multi-windows mode and subsurface mode
* Open 2 app windows
* Open the Waydroid full-ui window
* Crash

hwcomposer/wayland-hwc.cpp
```diff
@@ -772,12 +738,11 @@ keyboard_handle_enter(void *data, struct wl_keyboard *,
...
+    if (!window)
+        return; // Window was destroyed in hwc_set
 
-    if (window->display->task != nullptr) {
+    if (display->task != nullptr) {
         if (window->taskID != "none" && window->taskID != "0") {
```
aleasto, line 754: s/none/None/ to match what you're using in `split_layer_names_helper`
Actually this should probably be made resistant to non-integers and avoid stoi exceptions

hwcomposer/modes/waydroid_mode_impl.h
```diff
@@ -0,0 +1,139 @@
...
+
+            hwc_layer = skipped_layers.framebuffer_target_layer();
+            assert(hwc_layer);
+            i = skipped_layers.framebuffer_target_index();
+        } else if (skipped_layers.first_skipped() < i && i <= skipped_layers.last_skipped()) {
+            if (hwc_layer->acquireFenceFd != -1) {
+                close(hwc_layer->acquireFenceFd);
+            }
+            return 0;
+        } else if (hwc_layer == skipped_layers.framebuffer_target_layer() && skipped_layers.has_skipped_layers()) {
```
aleasto:
```suggestion
        } else if (hwc_layer == skipped_layers.framebuffer_target_layer()) {
```

In compositing mode we never want to render the framebuffer_target_layer, unless we have skipped layers that were composited into it, otherwise we're potentially overlaying stale buffers. Indeed with full-ui or single-app in subsurface compositing mode, I get a frozen image after the bootanimation (you might need to use shm buffers to reproduce)

hwcomposer/modes/waydroid_mode_impl.h
```diff
@@ -0,0 +1,139 @@
...
+            // The framebuffer_target layer will be drawn at i = skipped_layers.first_skipped()
+            // Skip it here
+            return 0;
+        } else if (hwc_layer->compositionType != HWC_OVERLAY && hwc_layer->compositionType != HWC_FRAMEBUFFER_TARGET) {
```
aleasto: Alternative to the above suggestion, this would achieve the same
```suggestion
        } else if (hwc_layer->compositionType != HWC_OVERLAY) {
```

hwcomposer/wayland-hwc.cpp
```diff
@@ -556,70 +464,61 @@ create_window(struct display *display, bool use_subsurfaces, std::string appID,
...
 
+    if (display->isMaximized || calibrating) {
+        window->set_maximize(true);
+    }
+    // Set title
+    const hidl_string appID_hidl(appID);
+    if (window->appID != "Waydroid" && display->task) {
+        display->task->getAppName(appID_hidl, [&](const hidl_string &value){
+            window->set_title(value.c_str());
+        });
```
aleasto:
```suggestion
    if (appID != "Waydroid" && display->task) {
        const hidl_string appID_hidl(appID);
        display->task->getAppName(appID_hidl, [&](const hidl_string &value){
            window->set_title(value.c_str());
        });
```

`window->appID` is unset here; I believe you meant to use the `appID` function argument.
Also `appID_hidl` can be defined inside the if scope.

hwcomposer/hwcomposer.cpp
aleasto: Mh I see this became `if (is_blacklisted(pdev, first_tid_layer->aid, ...)`

That's OK I think.

aleasto (changes requested):
OK I'm done for a first pass.

Left some in-line comments to be fixed

hwcomposer/wayland-hwc.cpp
```diff
@@ -460,47 +262,153 @@ struct wl_shell_surface_listener shell_surface_listener = {
...
+    } else {
+        assert(shell_surface);
+        wl_shell_surface_set_title(shell_surface, title);
+    }
+}
+
+void window::set_app_id(std::string appID) {
+    if (appID != "Waydroid")
+        appID = "waydroid." + appID;
+    this->appID = std::move(appID);
```
aleasto: This means that now `window->appID` has the `waydroid.` prefix and so cannot be matched against an android package name.

The `waydroid.` prefix is only needed in `xdg_toplevel_set_app_id` to match the desktop file.

See `WaydroidWindow.cpp` for places where we want to match a package name.

hwcomposer/modes/multi-window.cpp
```diff
@@ -0,0 +1,115 @@
...
+    return 0;
+}
+
+int multi_window_mode::cleanup_stale_windows(waydroid_hwc_composer_device_1* pdev) {
+    pdev->display->windows.erase_if([&](const auto& it){
+        const auto& key = it.first;
+        return std::all_of(layer_infos.container().begin(), layer_infos.container().end(), [&](const auto &layer_info){
+            return layer_info.type == LayerSplitType::Empty || layer_info.key() != key;
+        });
+    });
```
aleasto: See commit https://github.com/waydroid/android_hardware_waydroid/commit/0c48b2d8a97108214a36cd0d51b51f6972c5ce04

We need to filter layers that we later decide to not handle when checking if there's any visible layers for this window or not.

Perhaps it would be useful to abstract away a method that defines the set of layers that will be handled for the different waydroid modes?

aleasto:
> OK I'm done for a first pass.
> 
> Left some in-line comments to be fixed

Well I lied. Now I'm done. Maybe :)

hwcomposer/hwcomposer.cpp
Molytho: Yes, this should be fine in the latest version but I also fixed the earlier commit just in case

hwcomposer/hwcomposer.cpp
Molytho: Sounds reasonable. I will drop it for now

hwcomposer/wayland-hwc.cpp
Molytho: Ups. This change was unintentional. Changed it back to "none" in `split_layer_names_helper`

hwcomposer/modes/multi-window.cpp
Molytho: Is this only needed there?
What about those two cases?

https://github.com/waydroid/android_hardware_waydroid/blob/043435d70f41981905d26565371fd26751cdb157/hwcomposer/modes/single-window.cpp#L52-L55
https://github.com/waydroid/android_hardware_waydroid/blob/043435d70f41981905d26565371fd26751cdb157/hwcomposer/modes/single-window.cpp#L73-L75

hwcomposer/wayland-hwc.cpp
Molytho: Digging into the libwayland code, this does not (necessarily) sound like a KWin bug. libwayland validates that the compositor only sends null objects when it is allowed by the protocol, thus KWin sends an object id for an object that was already destroyed.
Either KWin sends this event when it does not yet know about the destruction of this object or it willingly sends this event for a destroyed surface (which would probably be a spec violation).
Either way we have to check for a NULL surface since the first case is valid.

aleasto: Yeah that makes sense. Thanks.

hwcomposer/modes/multi-window.cpp
aleasto: Indeed I believe this should apply to all three cases.

aleasto: I haven't verified this, but I think that this issue originates from the fact that surfaceflinger does not send the name of the HWC_FRAMEBUFFER_TARGET layer, so if the HWC_FRAMEBUFFER_TARGET layer is the last one in the list and the layer count decreases in between two commits, then the HWC_FRAMEBUFFER_TARGET layer will be assigned a stale name from the previous "last - 1" layer. So now we have a layer with a TID identifier but which doesn't actually belong to an android app.

hwcomposer/wayland-hwc.cpp
Molytho: I've added null checks for all wayland object parameters of wayland events (https://github.com/waydroid/android_hardware_waydroid/commit/26230b49ec99d48509518c5fccbc87a9a7d3f9fc).
Sadly this change isn't trivial so it might have unintended side effects in this corner case.

hwcomposer/modes/multi-window.cpp
Molytho: You are right. SurfaceFlinger does not write a layer name for the HWC_FRAMEBUFFER_TARGET layer.
Thinking about it, in single window mode we should just ignore the HWC_FRAMEBUFFER_TARGET layer since every other layer is shown on screen (including skipped layers).
I've added code to the split_layer_names_helper to ensure that the HWC_FRAMEBUFFER_TARGET layer gets an empty name.

aleasto: > I've added code to the split_layer_names_helper to ensure that the HWC_FRAMEBUFFER_TARGET layer gets an empty name.

Thanks. Sounds good.

hwcomposer/modes/multi-window.cpp
```diff
@@ -0,0 +1,123 @@
...
+}
+
+int multi_window_mode::cleanup_stale_windows(waydroid_hwc_composer_device_1* pdev,
+                                             hwc_display_contents_1_t* contents) {
+    pdev->display->windows.erase_if([&](const auto& it){
+        const auto& key = it.first;
+        for (size_t i = 0; i < contents->numHwLayers; ++i) {
+            const auto &layer_info = layer_infos[i];
+            const auto &layer = contents->hwLayers[i];
+            if (can_handle_layer_type(layer.compositionType)
```
aleasto: I'm not sure this is enough; What if all of its layers got marked with `HWC_SKIP_LAYER`?

I think this should match exactly the layers that we ignore in handle_layer (and thus be extracted into a function that both handle_layer and cleanup_stale_windows call).

hwcomposer/wayland-hwc.cpp
aleasto: >  Sadly this change isn't trivial so it might have unintended side effects in this corner case.

Seems fine to me, thank you.

hwcomposer/modes/multi-window.cpp
Molytho: Yeah, I already noticed this and I am working on it.
I was just debugging an issue which is introduced by this change.
Apparently there is a SurfaceFlinger bug where it resets the compositionType field without setting HWC_GEOMETRY_CHANGED.

Molytho:
I found a deadlock with https://github.com/Molytho/android_hardware_waydroid/commit/aca92c06f44891b57a7a550a8b86fdf207d16a8d.
Probably have to rethink the wayland thread again.

hwcomposer/modes/multi-window.cpp
Molytho: Should be done now

Molytho:
> I found a deadlock with [Molytho@aca92c0](https://github.com/Molytho/android_hardware_waydroid/commit/aca92c06f44891b57a7a550a8b86fdf207d16a8d). Probably have to rethink the wayland thread again.

I had to drop https://github.com/Molytho/android_hardware_waydroid/commit/aca92c06f44891b57a7a550a8b86fdf207d16a8d.
It appears that dispatching events only in the wayland thread is not so easy to achieve.
The details of the deadlock:
The xdg_surface_close event handler needs to lock windowsMutex thus the wayland thread needs to wait when hwc_set is currently executed, but window::create depends on wayland events being delivered thus this thread will also lock up. 
This is currently "not" an issue upstream since if the wayland thread locks up, the hwc_set thread will just dispatch the events it needs by itself.

To sum it up, although I believe that dispatching one wl_event_queue in two threads concurrently is probably a bad idea, it currently appears to work and is not trivially fixable thus I will leave it out of this PR.

aleasto (approved):
Looks good.

Pending some more testing on my side.

aleasto:
With the deferred window deletion, I'm seeing that closing the last window in multi-windows mode from the android side doesn't destroy the wayland window until the next pointer movement, presumably because once we stop committing surface updates the server doesn't have any event to send us so `wl_display_dispatch` doesn't return until some event like a mouse pointer move.

Can we somehow force `wl_display_dispatch` to return once? I see it has an optional timeout but that doesn't seem like a very good solution.

Molytho:
I feared this might happen.
There is no easy way to make `wl_display_dispatch` return early. We have to implement reading the wayland fd ourselves then. This way we can add an additional `event_fd` to the `poll` which we can use to wakeup the thread

aleasto:
Perhaps a better way to handle this is to manually ref count the window struct so that it can be kept alive in the wayland callbacks

Molytho:
I can't think of a way to make this work. For this to work the wayland objects (xdg_surface, xdg_toplevel, ...) need to hold a reference to struct window but there is no way for us to release this reference since we can't know when the wayland object is actually destroyed.
We also can't do this by setting the user_data of the objects to NULL when we want to destroy the window since the wayland thread might be in an event handler at this point which would still lead to a use-after-free.

Molytho:
I have come up with another idea but I think this deserves its own pull request because I want to make some further cleanup beforehand. I will drop the deferred delete commit for now. Since this bug also affects upstream it should (hopefully) not cause any regressions.

aleasto:
`waydroid show-full-ui` with `persist.waydroid.multi_windows=true` (which is not the intended way to use multi-windows, but because it works people use it) has some missing layers, because `hwc_prepare` sets all layers in-between the first and last skipped layers as `HWC_FRAMEBUFFER` only if `!pdev->multi_windows`, but then we still use the compositing-full-ui mode which skips all layers in between. Somehow this worked correctly before.

This is particularly evident in Android 13 where every floating-window gets a background layer marked as SKIP (why?), so the layers look like this:
* `systemui` `HWC_OVERLAY`
* `foo/Background` `HWC_FRAMEBUFFER` `SKIP`
* `foo/MainSurface` `HWC_OVERLAY`
* `bar/Background` `HWC_FRAMEBUFFER` `SKIP`
* `bar/MainSurface` `HWC_OVERLAY`

and we draw
* `systemui`
* `HWC_FRAMEBUFFER_TARGET` (backgorund of foo and bar)
* `bar/MainSurface`

so `foo/MainSurface` never gets drawn.

Molytho:
> `waydroid show-full-ui` with `persist.waydroid.multi_windows=true` (which is not the intended way to use multi-windows, but because it works people use it) has some missing layers, because `hwc_prepare` sets all layers in-between the first and last skipped layers as `HWC_FRAMEBUFFER` only if `!pdev->multi_windows`, but then we still use the compositing-full-ui mode which skips all layers in between. Somehow this worked correctly before.
> 
> This is particularly evident in Android 13 where every floating-window gets a background layer marked as SKIP (why?), so the layers look like this:
> 
>     * `systemui` `HWC_OVERLAY`
> 
>     * `foo/Background` `HWC_FRAMEBUFFER` `SKIP`
> 
>     * `foo/MainSurface` `HWC_OVERLAY`
> 
>     * `bar/Background` `HWC_FRAMEBUFFER` `SKIP`
> 
>     * `bar/MainSurface` `HWC_OVERLAY`
> 
> 
> and we draw
> 
>     * `systemui`
> 
>     * `HWC_FRAMEBUFFER_TARGET` (backgorund of foo and bar)
> 
>     * `bar/MainSurface`
> 
> 
> so `foo/MainSurface` never gets drawn.

This should be fixed by the newest commit. Now hwc_prepare should stick to full-ui mode behaviour in this case.

hwcomposer/modes/waydroid_mode.h
```diff
@@ -0,0 +1,106 @@
...
+enum class LayerSplitType {
+    TID,
+    RawName,
+    Empty
+};
+struct layer_info {
+    LayerSplitType type;
+    std::string aid;
+    std::string tid;
+    std::string components {};
```
aleasto: `component` singular please, one layer has one AID and one component

hwcomposer/modes/multi-window.cpp
```diff
@@ -0,0 +1,126 @@
...
+    pdev->display->ignored_apps.clear();
+
+    return 0;
+}
+
+int multi_window_mode::handle_layer(waydroid_hwc_composer_device_1* pdev, hwc_layer_1* hwc_layer, size_t i) {
+    if (!can_handle_layer(*hwc_layer)) {
+        if (hwc_layer->compositionType != HWC_FRAMEBUFFER_TARGET) {
+            // TODO: Support HWC_BACKGROUND
+            ALOGW("Encountered unsupported compositionType: %" PRId32, hwc_layer->compositionType);
```
aleasto, line 100: Let's avoid printing this when it's a skip layer: we already know we cannot handle skip layers in multi-window mode. This is otherwise spammy because skip layers are kept as HWC_FRAMEBUFFER and in Android 13 we get one skip layer with component="Background" per every task.

aleasto (commented):
Sorry, busy week.

Found a couple more nits.

I also still need to test this for regressions on different compositors and manually disabling "optional" protocols.

hwcomposer/wayland-hwc.cpp
```diff
@@ -460,47 +261,158 @@ struct wl_shell_surface_listener shell_surface_listener = {
...
+        xdg_toplevel_set_minimized(xdg_toplevel);
+        wl_surface_commit(surface); // unclear if this is required
+        wl_display_flush(display->display);
+    }
+}
+
+window::~window() {
+    layers.clear();
+    if (xdg_toplevel)
+        xdg_toplevel_destroy(xdg_toplevel);
```
aleasto: This sequence generates events Wayland events in an invalid order:
> When a client wants to destroy a wl_surface, they must destroy this 'role object' before the wl_surface.

If our window has no background (e.g. because there's no `wl_subcompositor`), the `wl_surface` for this toplevel is `layer[0]` so it will be destroyed in `layers.clear()` before the toplevel role object is destroyed. Some compositors don't seem to complain, but Mutter does.

hwcomposer/gralloc_handler.h
```diff
@@ -0,0 +1,68 @@
...
+ * NONINFRINGEMENT.  IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+ * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#pragma once
+
+#include <asm/unistd_64.h>
```
aleasto: I guess this is for `__NR_memfd_create`. But this path only exists on x86; the generic `unistd.h` should be fine.

It also seems unused like a few other headers in this `gralloc_handler.h`, so move it to the cpp

aleasto: Mh actually it seems to be defined in `syscall.h`

hwcomposer/gralloc_handler.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+void update_shm_buffer_generic(display *display, buffer *buffer) {
+    display->egl_work_queue.emplace_back(std::bind(egl_render_to_pixels, display, buffer));
+    sem_post(&display->egl_go);
+    sem_wait(&display->egl_done);
+}
+void update_shm_buffer_default(display *, buffer *buffer) {
+    void *data;
+    uint32_t shm_stride, src_stride;
+    android::Rect bounds(buffer->metadata.width, buffer->metadata.height);
```
aleasto, line 281: `#include <ui/Rect.h>` explicitely in case it's not recursively included (it's not in older AOSP, I was testing on lineage-16.0 now)

aleasto:
The window snapshotting functionality is not working properly anymore; it seems to be snapshotting the first frame of the new foregroud task rather than the last frame of the previous foregroud task.

Molytho:
> The window snapshotting functionality is not working properly anymore; it seems to be snapshotting the first frame of the new foregroud task rather than the last frame of the previous foregroud task.

Should work now. I broke the whole snapshotting functionality by clearing the last_layer_buffer field before trying the actual snapshotting.

aleasto:
I have one big remaining issue, and one small one.

The small one: I've been seeing crashes in `dmabuf_format`, presumably because it's being called from different threads simultaneously. If the same threading issue existed in lineage-18.1 already then perhaps the move to std::unordered_set has made it easier to hit. I'm wondering if we should add a small mutex as a temporary measure. It's not a big deal because after crashing hwcomposer will just try starting again and most likely succeed the second time.

The big one: Ubuntu Touch using Lomiri/Mir 1.x is now broken in full-ui mode (and probably all other modes too). Doesn't show any window. I do not have a working Ubuntu Touch device at the moment so I'll have to try and run Mir 1.x on my computer, but I've asked a member of our community to test on their Ubuntu Touch and that was the result. Mir 1.x is quite limited in its Wayland conformance, but it used to work and we've got a "large" user-base there.

Molytho:
I am aware of the first one and pushed a workaround that hopefully makes it much less likely (assuming that the formats do not change this should actually "fix" it)

Regarding the second one: Yeah this sounds bad. It is very possible that I broke some legacy stuff and I am probably unable to debug this properly since I got no Ubuntu Touch device.

JamiKettunen:
https://github.com/ubports/ubports-pdk should (almost) replicate a normal Ubuntu Touch device including Mir 1.x on any desktop with QEMU/KVM and read-write filesystem for easier hacking around

aleasto:
Ok, I managed to get an Ubuntu Touch replica environment.

Here's the bug: since commit "[hwcomposer: Simplify struct window](https://github.com/waydroid/android_hardware_waydroid/pull/63/commits/077db2cbfe5db44ed11915f55acec5526fb7f87b)" we are now always sending a black background surface, even in non-compositing mode, and Lomiri/Mir1.x don't like that. It might be because it does not implement wp_viewporter and gets confused at a 1x1 main surface with a larger subsurface on top (it's allowed by the spec, but...), or it might be because subsurfaces are kinda broken in general.

I think that we should add `!display->viewporter` to the conditions that skip the window background logic, which both makes sense as a 1x1 buffer is not a background, and helps catch the broken Ubuntu Touch compositor.

While at it, I've found while bisecting that the commit "[hwcomposer: Replace surfaces, subsurfaces and viewports in struct window with one map and a helper struct](https://github.com/waydroid/android_hardware_waydroid/pull/63/commits/79d39dd85e60f7cc05b57299bf7552426b61b3be)" does not compile. I really appreciate how you've generally kept every commit buildable, so please fix this one as well :)

hwcomposer/wayland-hwc.cpp
```diff
@@ -634,57 +535,125 @@ create_window(struct display *display, bool use_subsurfaces, std::string appID,
...
+    }
+
+    // TODO: Fix background when viewport is not supported
     // No subsurface background for us!
-    if ((!use_subsurfaces && !display->subcompositor) ||
+    if (!display->subcompositor ||
+        !display->viewporter ||
         property_get_bool("persist.waydroid.no_background_subsurface", false)) {
+        window->destroy_background_objects = false;
+        window->layers.emplace_back(window->surface, window->viewport);
```
aleasto, line 563: `window->viewport` is now always NULL here, but we still want a viewport if we were entering this branch just because of `!display->subcompositor` or the prop

Molytho:
> Here's the bug: since commit "[hwcomposer: Simplify struct window](https://github.com/waydroid/android_hardware_waydroid/pull/63/commits/077db2cbfe5db44ed11915f55acec5526fb7f87b)" we are now always sending a black background surface, even in non-compositing mode, and Lomiri/Mir1.x don't like that. It might be because it does not implement wp_viewporter and gets confused at a 1x1 main surface with a larger subsurface on top (it's allowed by the spec, but...), or it might be because subsurfaces are kinda broken in general.

May it be possible that Lomiri/Mir1.x is using the [wl_shell](https://wayland.app/protocols/wayland#wl_shell) protocol instead of [xdg_wm_base](https://wayland.app/protocols/xdg-shell#xdg_wm_base)? Then this would make sense to me since wl_shell_surface has no set_window_geometry and likely just assumes that the size of the main surface is the window size.

> I think that we should add `!display->viewporter` to the conditions that skip the window background logic, which both makes sense as a 1x1 buffer is not a background, and helps catch the broken Ubuntu Touch compositor.

This is a behaviour change but as you said, a 1x1 buffer is not really a background so I also think this is reasonable. Should be done now.

> While at it, I've found while bisecting that the commit "[hwcomposer: Replace surfaces, subsurfaces and viewports in struct window with one map and a helper struct](https://github.com/waydroid/android_hardware_waydroid/pull/63/commits/79d39dd85e60f7cc05b57299bf7552426b61b3be)" does not compile. I really appreciate how you've generally kept every commit buildable, so please fix this one as well :)

Done

hwcomposer/wayland-hwc.cpp
Molytho: Ups. Yeah true. Be right back

aleasto:
> May it be possible that Lomiri/Mir1.x is using the [wl_shell](https://wayland.app/protocols/wayland#wl_shell) protocol instead of [xdg_wm_base](https://wayland.app/protocols/xdg-shell#xdg_wm_base)? Then this would make sense to me since wl_shell_surface has no set_window_geometry and likely just assumes that the size of the main surface is the window size.

No it is using xdg-shell :shrug:

aleasto (approved):

hwcomposer/gralloc_handler.cpp
```diff
@@ -0,0 +1,339 @@
...
+    int *the_ints = (int *)wl_array_add(&ints, handle->numInts * sizeof(int));
+    memcpy(the_ints, handle->data + handle->numFds, handle->numInts * sizeof(int));
+    android_wlegl_handle *wlegl_handle = android_wlegl_create_handle(display->android_wlegl, handle->numFds, &ints);
+    wl_array_release(&ints);
+
+    for (int i = 0; i < handle->numFds; i++) {
+        android_wlegl_handle_add_fd(wlegl_handle, handle->data[i]);
+    }
+
+    buf->wl_buffer = android_wlegl_create_buffer(display->android_wlegl, buf->metadata.height, buf->metadata.height, buf->metadata.pixel_stride, metadata.format, GRALLOC_USAGE_HW_RENDER, wlegl_handle);
```
aleasto: buf->metadata.width :)

aleasto (changes requested):

aleasto:
Alright let's do this!

PR #68, opened 2026-01-04
hwcomposer: Blacklist com.google.android.tvlauncher for ATV

### LINKED ISSUE #62: Home screen / black screen when exiting apps
Hi, I just want to use this build of Android TV for launching two TV apps (Spotify, Symfonium) on a Linux host that are otherwise remote-unfriendly (app icons #​3 and ​#​4 in the screenshots below).

I launch the desired apps via KDE Plasma using its panel and that works wonderfully... but doing so also opens the Android TV Home screen in another Waydroid window at the same time. Closing the desired app goes back to the Android TV Home screen every time, and trying to close the Android TV Home screen with Alt + F4 (as I do with apps) just immediately re-opens the Home screen.

The only way to get the Android TV Home screen to go away is to stop the Waydroid session or kill Waydroid entirely. I can't find any way to minimize or hide it, and having it open impacts performance in host OS apps so I definitely do not want to leave any Android TV windows visible.

I tried the following from [this Reddit thread](https://www.reddit.com/r/ShieldAndroidTV/comments/o96npc/instructions_for_replacing_android_tv_home_with_a/):

> 6. Now issue the following command to disable Android TV Home:
> 
>     `adb shell pm disable-user --user 0 com.google.android.tvlauncher`
> 
> 7. You can safely disable Android TV Core:
> 
>     `adb shell pm disable-user --user 0 com.google.android.tvrecommendations`

This didn't change the behavior whatsoever, it just makes a black Waydroid screen open every time instead. As a result, it's all black in the images below - but otherwise, it's the Android TV Home screen, or whatever home launcher I set otherwise (tested with Projectivy)

Screenshot showing the additional Waydroid window that opens alongside apps:
Image

Screenshot showing the window in action:
Image

If possible, I'd like to exclusively launch these 2 TV apps (Spotify, Symfonium) via KDE Plasma, and cleanly close them as desired. I couldn't care less about CPU or graphics performance when these desired apps are open, since they're exclusively used for playing audio.

I wouldn't necessarily mind running Waydroid full screen and using all my apps through it, but...
- If I run Waydroid on top of the existing KDE Plasma, that impacts performance of host apps, which is a problem when watching content via YouTube or Emby.
- If I run Waydroid exclusively without a DE and instead use the Android TV versions of my host OS apps, I will (as far as I'm aware) lose the benefits of running a Linux host - no KDE Plasma follow-the-sun Night Light mode that respects seasons throughout the year (Twilight on Android TV is no match for this), no software compressor (I don't know of any Android TV alternative). At that point, an Nvidia Shield or equivalent would be an almost identical experience.

Is this a limitation of Android TV and/or the build in question? Is there any way to entirely prohibit this behavior somehow? Thanks in advance!

supechicken:
This fixes https://github.com/supechicken/waydroid-androidtv-build/issues/62

> I launch the desired apps via KDE Plasma using its panel and that works wonderfully... but doing so also opens the Android TV Home screen in another Waydroid window at the same time. Closing the desired app goes back to the Android TV Home screen every time, and trying to close the Android TV Home screen with Alt + F4 (as I do with apps) just immediately re-opens the Home screen.

### DIFF
--- a/hwcomposer/hwcomposer.cpp
+++ b/hwcomposer/hwcomposer.cpp
@@ -727,6 +727,8 @@ static int hwc_open(const struct hw_module_t* module, const char* name,
 
     pdev->blacklisted_apps["com.android.launcher3"] = {};
     pdev->blacklisted_apps["com.android.settings"] = {"com.android.settings.FallbackHome"};
+    pdev->blacklisted_apps["com.android.tv.settings"] = {"com.android.tv.settings.system.FallbackHome"};
+    pdev->blacklisted_apps["com.google.android.tvlauncher"] = {};
     if (property_get("waydroid.blacklist_apps", property, nullptr) > 0) {
         std::string blacklist_apps = std::string(property);
         std::istringstream iss(blacklist_apps);


aleasto:
Thanks

PR #69, opened 2026-03-21
vulkan: add ETC2/EAC format compatibility layer for Intel GPUs

### LINKED ISSUE #702: Most games don't work
**Describe the bug**
I use the official Waydroid image of GApps LineageOS 18.1 and I installed libndk via waydroid-script. But most games I tried crash when I try to play them. It doesn't matter if I use libndk or libhoudini, as it acts the same. I tried it on my laptop with i5-8250U and UHD 620 and there were the same issues. Some games worked like Geometry Dash or Bad Piggies, but others like Angry Birds 2, Konosuba and Blue Archive did not.

**Screenshots**
Konosuba game not working

https://user-images.githubusercontent.com/97224849/213174973-20cde39d-3faf-4fd0-af66-f0815cf635b6.mp4

Blue Archive not working

https://user-images.githubusercontent.com/97224849/213175038-ab9e38b8-b10f-42e1-b608-b44ea9e78eb3.mp4

**General information (please complete the following information):**
 - Waydroid tools Version [e.g. 1.3.4]
 - Waydroid Images Version [lineage-18.1-20230107-GAPPS]

**Desktop (please complete the following information):**
 - OS: [Arch Linux x86_64]
 - GPU: [e.g. AMD Radeon™ RX 6700XT]
 - Kernel version: [6.1.6-zen1-1-zen]
 - Host mesa version: [mesa-git 23.0.0_devel]
 - Desktop: [Hyprland]

**Additional context**

**Logs (please upload as file)**
 - Prop file, located on `/var/lib/waydroid/waydroid_base.prop`
sys.use_memfd=true
debug.stagefright.ccodec=0
ro.hardware.gralloc=gbm
ro.hardware.egl=mesa
ro.hardware.vulkan=radeon
ro.hardware.camera=v4l2
ro.opengles.version=196609
waydroid.system_ota=https://ota.waydro.id/system/lineage/waydroid_x86_64/GAPPS.json
waydroid.vendor_ota=https://ota.waydro.id/vendor/waydroid_x86_64/MAINLINE.json
waydroid.tools_version=1.3.4
ro.vndk.lite=true
waydroid.host.user=valorix
waydroid.host.uid=1000
waydroid.host.gid=1000
waydroid.xdg_runtime_dir=/run/user/1000
waydroid.pulse_runtime_path=/run/user/1000/pulse
waydroid.wayland_display=wayland-1
waydroid.stub_sensors_hal=1
 - Output of `waydroid log`
 - *Full* (not partially) Output of `waydroid logcat` from the beginning of starting waydroid services

### LINKED ISSUE #2187: [BUG] Intel gpu causes crash of waydroid
### Describe the bug

when i run it it pops up a black screen then just crash, i need to use software rendering when i want to use it but i want hardware acceleration

### Waydroid version

1.6.0

### Device

Linux Desktop

### Operating System

Nobara Linux 43

### Kernel version

6.18.2-200.nobara.fc43.x86_64

### Desktop Environment

GNOME 49

### GPU

Nvidia RTX 5060 (dgpu) Intel Raptor Lake-S UHD Graphics (thats what linux tells me)

### Logs

[waydroid-bugreport.tar.xz.gz](https://github.com/user-attachments/files/24546347/waydroid-bugreport.tar.xz.gz)

uzairdeveloper223:
### Problem

ARM64 Unity games crash with SIGABRT when running on Waydroid with Intel GPUs. The crash originates in `AHardwareBuffer_getNativeHandle` because Mesa ANV does not handle `AHardwareBuffer` with ETC2 compressed texture formats through gralloc. When gralloc encounters unsupported format `0x38` (`ETC2_RGB8`), the Vulkan swapchain creation path reaches `AHardwareBuffer_getNativeHandle` with an invalid handle and aborts.

The crash chain is: `UnityMain` calls `vkCreateSwapchainKHR`, which goes through `vulkan::driver::CreateSwapchainKHR` into `vulkan.intel.so`, where `AHardwareBuffer_getNativeHandle` receives a null native handle and triggers a SIGABRT.

### Solution

This adds a Vulkan implicit layer (`VK_LAYER_WAYDROID_compat`) that intercepts `vkGetPhysicalDeviceFormatProperties` and `vkGetPhysicalDeviceFormatProperties2`. For all ETC2/EAC formats, the layer reports zero feature flags (`linearTilingFeatures`, `optimalTilingFeatures`, `bufferFeatures`), which causes apps to detect that ETC2 is not supported and fall back to uncompressed textures. The layer also walks the `pNext` chain to zero out `VkFormatProperties3` when present, ensuring complete coverage for Vulkan 1.3 queries.

Dispatch tables are stored per-instance using a linked list, so the layer correctly handles multiple `VkInstance` objects existing simultaneously (for example from native services alongside applications). All linked list access is protected by a `pthread_mutex_t` for thread safety. The physical device to instance mapping uses the Vulkan loader dispatch key, which is the same approach used by the Khronos validation layers.

An init script automatically sets `debug.vulkan.layers=VK_LAYER_WAYDROID_compat` when `ro.hardware.vulkan=intel` is detected, so no user configuration is required.

### Activation scope (Intel only)

The layer implementation itself is GPU agnostic, but the init script only activates it when `ro.hardware.vulkan=intel` is detected. Other GPU vendors are unaffected at runtime. If someone were to manually enable the layer on AMD or another GPU via `setprop`, it would still function correctly by masking ETC2 formats, but there is no reason to do so since other drivers handle ETC2 through AHardwareBuffer without crashing.

### Format mappings

The layer masks the following ten ETC2/EAC formats. These are the baseline compressed texture formats for Android since OpenGL ES 3.0 and are primarily used by games built for ARM GPUs. Apps that target desktop Vulkan typically do not request ETC2 swapchain images. Games built for ARM with ETC2 as the only texture format will fall back to uncompressed RGBA when the layer is active.

| Format | Vulkan enum |
|---|---|
| ETC2 RGB8 UNORM | `VK_FORMAT_ETC2_R8G8B8_UNORM_BLOCK` |
| ETC2 RGB8 SRGB | `VK_FORMAT_ETC2_R8G8B8_SRGB_BLOCK` |
| ETC2 RGBA1 UNORM | `VK_FORMAT_ETC2_R8G8B8A1_UNORM_BLOCK` |
| ETC2 RGBA1 SRGB | `VK_FORMAT_ETC2_R8G8B8A1_SRGB_BLOCK` |
| ETC2 RGBA8 UNORM | `VK_FORMAT_ETC2_R8G8B8A8_UNORM_BLOCK` |
| ETC2 RGBA8 SRGB | `VK_FORMAT_ETC2_R8G8B8A8_SRGB_BLOCK` |
| EAC R11 UNORM | `VK_FORMAT_EAC_R11_UNORM_BLOCK` |
| EAC R11 SNORM | `VK_FORMAT_EAC_R11_SNORM_BLOCK` |
| EAC RG11 UNORM | `VK_FORMAT_EAC_R11G11_UNORM_BLOCK` |
| EAC RG11 SNORM | `VK_FORMAT_EAC_R11G11_SNORM_BLOCK` |

### Files

| File | Description |
|------|-------------|
| [vulkan/Android.bp](https://github.com/uzairdeveloper223/android_hardware_waydroid/blob/lineage-20/vulkan/Android.bp) | Soong build rules for the shared library and JSON manifest |
| [vulkan/VkLayer_waydroid_compat.c](https://github.com/uzairdeveloper223/android_hardware_waydroid/blob/lineage-20/vulkan/VkLayer_waydroid_compat.c) | Layer implementation with per-instance dispatch and thread safety |
| [vulkan/VkLayer_waydroid_compat.json](https://github.com/uzairdeveloper223/android_hardware_waydroid/blob/lineage-20/vulkan/VkLayer_waydroid_compat.json) | Layer manifest for the Vulkan loader |
| [vulkan/init.vulkan.waydroid.rc](https://github.com/uzairdeveloper223/android_hardware_waydroid/blob/lineage-20/vulkan/init.vulkan.waydroid.rc) | Init script that auto enables the layer on Intel GPUs |

### Validation done

Tested on Intel i5 8th gen with UHD 620, Ubuntu 24.04, Waydroid with LineageOS 20. Before the fix, ARM64 Unity games (com.MA.LAC) crashed immediately with SIGABRT on launch. After the fix, games launch and run without crashing. Textures fall back to RGBA uncompressed formats. Frame rate and visual output were identical to a reference run with ETC2 disabled in the Unity build settings. No rendering artifacts observed. Non Unity apps and non ETC2 Vulkan workloads are completely unaffected because the layer only modifies queries for the ten formats listed above.

Multi-instance dispatch was verified on the host by compiling the layer and running a test that creates two VkInstance objects, queries ETC2 format properties from the first instance's physical device, and destroys both instances. ETC2 features were correctly zeroed and both instances were destroyed without errors.

```
Creating instance 1...
Creating instance 2...
ETC2 RGB8 features: linear=0 optimal=0 buffer=0
PASS: ETC2 correctly masked
PASS: Both instances destroyed cleanly
```

### Known limitations

The layer does not perform actual ETC2 decompression. It hides format support so that apps use alternative uncompressed formats. This means slightly higher memory usage for textures compared to native ETC2 support, but has no visible impact on rendering quality or frame rate based on testing.

### Why not fix in Mesa?

The correct long term fix is to either add a null check guard in get_ahb_buffer_format_properties2() (prevents the crash) or wire ETC2 through anv_get_compressed_format_emulation() using the existing `vk_texcompress_etc2.h` infrastructure (adds actual support). A Mesa patch for the null check guard has been submitted separately at [waydroid/android_external_mesa3d#12](https://github.com/waydroid/android_external_mesa3d/pull/12). This layer remains useful as an immediate Waydroid side workaround that provides relief for affected users regardless of which Mesa version is deployed.

Fixes waydroid/waydroid#702
Fixes waydroid/waydroid#2187

Signed-off-by: Uzair Mughal <contact@uzair.is-a.dev>

### DIFF
--- a/vulkan/VkLayer_waydroid_compat.c
+++ b/vulkan/VkLayer_waydroid_compat.c
@@ -26,6 +26,9 @@
  * This layer intercepts vkGetPhysicalDeviceFormatProperties{,2} and
  * reports zero feature flags for all ETC2/EAC formats, causing apps
  * to fall back to uncompressed textures.
+ *
+ * Note: uses a single global dispatch table. This is safe for Waydroid
+ * since it only ever creates a single Vulkan instance per container.
  */
 
 #include <string.h>
@@ -33,7 +36,6 @@
 #include <vulkan/vulkan.h>
 #include <vulkan/vk_layer.h>
 
-/* Dispatch table, single-instance (Waydroid only creates one) */
 typedef struct {
     PFN_vkGetInstanceProcAddr GetInstanceProcAddr;
     PFN_vkDestroyInstance DestroyInstance;
@@ -64,7 +66,29 @@ static int is_etc2_eac_format(VkFormat format)
     }
 }
 
-/* Intercepted: zero out features for ETC2/EAC formats */
+static void zero_format_features(VkFormatProperties *props)
+{
+    props->linearTilingFeatures = 0;
+    props->optimalTilingFeatures = 0;
+    props->bufferFeatures = 0;
+}
+
+/* Walk the pNext chain and zero out VkFormatProperties3 if present */
+static void zero_format_features3(VkFormatProperties2 *pFormatProperties)
+{
+    VkBaseOutStructure *ext = (VkBaseOutStructure *)pFormatProperties->pNext;
+    while (ext) {
+        if (ext->sType == VK_STRUCTURE_TYPE_FORMAT_PROPERTIES_3) {
+            VkFormatProperties3 *props3 = (VkFormatProperties3 *)ext;
+            props3->linearTilingFeatures = 0;
+            props3->optimalTilingFeatures = 0;
+            props3->bufferFeatures = 0;
+            break;
+        }
+        ext = ext->pNext;
+    }
+}
+
 static VKAPI_ATTR void VKAPI_CALL
 compat_GetPhysicalDeviceFormatProperties(VkPhysicalDevice physicalDevice,
                                          VkFormat format,
@@ -72,11 +96,8 @@ compat_GetPhysicalDeviceFormatProperties(VkPhysicalDevice physicalDevice,
 {
     g_dispatch.GetPhysicalDeviceFormatProperties(physicalDevice, format,
                                                  pFormatProperties);
-    if (is_etc2_eac_format(format)) {
-        pFormatProperties->linearTilingFeatures = 0;
-        pFormatProperties->optimalTilingFeatures = 0;
-        pFormatProperties->bufferFeatures = 0;
-    }
+    if (is_etc2_eac_format(format))
+        zero_format_features(pFormatProperties);
 }
 
 static VKAPI_ATTR void VKAPI_CALL
@@ -87,9 +108,8 @@ compat_GetPhysicalDeviceFormatProperties2(VkPhysicalDevice physicalDevice,
     g_dispatch.GetPhysicalDeviceFormatProperties2(physicalDevice, format,
                                                   pFormatProperties);
     if (is_etc2_eac_format(format)) {
-        pFormatProperties->formatProperties.linearTilingFeatures = 0;
-        pFormatProperties->formatProperties.optimalTilingFeatures = 0;
-        pFormatProperties->formatProperties.bufferFeatures = 0;
+        zero_format_features(&pFormatProperties->formatProperties);
+        zero_format_features3(pFormatProperties);
     }
 }
 
--- a/vulkan/VkLayer_waydroid_compat.c
+++ b/vulkan/VkLayer_waydroid_compat.c
@@ -1,0 +1,241 @@
+/*
+ * Copyright (C) 2026 The Waydroid Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * Vulkan implicit layer that masks ETC2/EAC format support to prevent
+ * crashes on Intel GPUs when ARM64 apps use AHardwareBuffer with
+ * compressed textures.
+ *
+ * Mesa ANV does not handle AHardwareBuffer with ETC2 formats through
+ * gralloc, causing a SIGABRT in AHardwareBuffer_getNativeHandle when
+ * gralloc encounters unsupported format 0x38 (ETC2_RGB8).
+ *
+ * This layer intercepts vkGetPhysicalDeviceFormatProperties{,2} and
+ * reports zero feature flags for all ETC2/EAC formats, causing apps
+ * to fall back to uncompressed textures.
+ */
+
+#include <string.h>
+#include <stdlib.h>
+#include <vulkan/vulkan.h>
+#include <vulkan/vk_layer.h>
+
+/* Dispatch table, single-instance (Waydroid only creates one) */
+typedef struct {
+    PFN_vkGetInstanceProcAddr GetInstanceProcAddr;
+    PFN_vkDestroyInstance DestroyInstance;
+    PFN_vkEnumeratePhysicalDevices EnumeratePhysicalDevices;
+    PFN_vkGetPhysicalDeviceFormatProperties GetPhysicalDeviceFormatProperties;
+    PFN_vkGetPhysicalDeviceFormatProperties2 GetPhysicalDeviceFormatProperties2;
+    PFN_vkEnumerateDeviceExtensionProperties EnumerateDeviceExtensionProperties;
+} InstanceDispatch;
+
+static InstanceDispatch g_dispatch;
+
+static int is_etc2_eac_format(VkFormat format)
+{
+    switch (format) {
+    case VK_FORMAT_ETC2_R8G8B8_UNORM_BLOCK:
+    case VK_FORMAT_ETC2_R8G8B8_SRGB_BLOCK:
+    case VK_FORMAT_ETC2_R8G8B8A1_UNORM_BLOCK:
+    case VK_FORMAT_ETC2_R8G8B8A1_SRGB_BLOCK:
+    case VK_FORMAT_ETC2_R8G8B8A8_UNORM_BLOCK:
+    case VK_FORMAT_ETC2_R8G8B8A8_SRGB_BLOCK:
+    case VK_FORMAT_EAC_R11_UNORM_BLOCK:
+    case VK_FORMAT_EAC_R11_SNORM_BLOCK:
+    case VK_FORMAT_EAC_R11G11_UNORM_BLOCK:
+    case VK_FORMAT_EAC_R11G11_SNORM_BLOCK:
+        return 1;
+    default:
+        return 0;
+    }
+}
+
+/* Intercepted: zero out features for ETC2/EAC formats */
+static VKAPI_ATTR void VKAPI_CALL
+compat_GetPhysicalDeviceFormatProperties(VkPhysicalDevice physicalDevice,
+                                         VkFormat format,
+                                         VkFormatProperties *pFormatProperties)
+{
+    g_dispatch.GetPhysicalDeviceFormatProperties(physicalDevice, format,
+                                                 pFormatProperties);
+    if (is_etc2_eac_format(format)) {
+        pFormatProperties->linearTilingFeatures = 0;
+        pFormatProperties->optimalTilingFeatures = 0;
+        pFormatProperties->bufferFeatures = 0;
+    }
+}
+
+static VKAPI_ATTR void VKAPI_CALL
+compat_GetPhysicalDeviceFormatProperties2(VkPhysicalDevice physicalDevice,
+                                          VkFormat format,
+                                          VkFormatProperties2 *pFormatProperties)
+{
+    g_dispatch.GetPhysicalDeviceFormatProperties2(physicalDevice, format,
+                                                  pFormatProperties);
+    if (is_etc2_eac_format(format)) {
+        pFormatProperties->formatProperties.linearTilingFeatures = 0;
+        pFormatProperties->formatProperties.optimalTilingFeatures = 0;
+        pFormatProperties->formatProperties.bufferFeatures = 0;
+    }
+}
+
+/* Instance creation: set up dispatch chain */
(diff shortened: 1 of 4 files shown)


vulkan/VkLayer_waydroid_compat.c
```diff
@@ -0,0 +1,241 @@
...
+    case VK_FORMAT_ETC2_R8G8B8A8_SRGB_BLOCK:
+    case VK_FORMAT_EAC_R11_UNORM_BLOCK:
+    case VK_FORMAT_EAC_R11_SNORM_BLOCK:
+    case VK_FORMAT_EAC_R11G11_UNORM_BLOCK:
+    case VK_FORMAT_EAC_R11G11_SNORM_BLOCK:
+        return 1;
+    default:
+        return 0;
+    }
+}
```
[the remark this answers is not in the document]
aleasto: boring

vulkan/VkLayer_waydroid_compat.c
```diff
@@ -0,0 +1,241 @@
...
+    return NULL;
+}
+
+static VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL
+compat_GetDeviceProcAddr(VkDevice device, const char *pName)
+{
+    (void)device;
+    (void)pName;
+    return NULL;
+}
```
aleasto: Why is this defined this way instead of setting `pVersionStruct->pfnGetDeviceProcAddr = NULL` and letting the loader skip to the next layer?

vulkan/VkLayer_waydroid_compat.c
```diff
@@ -0,0 +1,241 @@
...
+                                          VkFormat format,
+                                          VkFormatProperties2 *pFormatProperties)
+{
+    g_dispatch.GetPhysicalDeviceFormatProperties2(physicalDevice, format,
+                                                  pFormatProperties);
+    if (is_etc2_eac_format(format)) {
+        pFormatProperties->formatProperties.linearTilingFeatures = 0;
+        pFormatProperties->formatProperties.optimalTilingFeatures = 0;
+        pFormatProperties->formatProperties.bufferFeatures = 0;
+    }
```
aleasto, line 187: /gemini do we also need to handle VkFormatProperties3 from `pFormatProperties->pNext`?

uzairdeveloper223:
@aleasto Done!

aleasto:
@uzairdeveloper223 there's still 2 unresolved in-line comments

uzairdeveloper223:
> @uzairdeveloper223 there's still 2 unresolved in-line comments

Addressed both @aleasto, the VkFormatProperties3 pNext is handled by zero_format_features3(), and removed the compat_GetDeviceProcAddr stub in favor of pfnGetDeviceProcAddr = NULL. Updated in the latest push

vulkan/VkLayer_waydroid_compat.c
```diff
@@ -0,0 +1,253 @@
...
+ * Mesa ANV does not handle AHardwareBuffer with ETC2 formats through
+ * gralloc, causing a SIGABRT in AHardwareBuffer_getNativeHandle when
+ * gralloc encounters unsupported format 0x38 (ETC2_RGB8).
+ *
+ * This layer intercepts vkGetPhysicalDeviceFormatProperties{,2} and
+ * reports zero feature flags for all ETC2/EAC formats, causing apps
+ * to fall back to uncompressed textures.
+ *
+ * Note: uses a single global dispatch table. This is safe for Waydroid
+ * since it only ever creates a single Vulkan instance per container.
```
aleasto: I'm still not sure about this. It might be the case for applications, but I imagine there's no guarantee for native services linking against libvulkan.

uzairdeveloper223: Hmm yeah that's a fair point actually. I was thinking about it from the app side where Waydroid only runs one app at a time, but you're right that native services linking against libvulkan could also end up creating instances and the global table would get clobbered. I hadn't considered that path.

I can rework it to use a per-instance dispatch table if you think it's worth doing for this. It would mean adding a small lookup by VkInstance handle in CreateInstance/DestroyInstance and having the intercepted functions resolve their dispatch from the physical device's parent instance. Not a huge change but it does add some complexity.

On another note, I've also been looking at where this actually breaks in Mesa. The root cause is in vk_android.c where vk_ahb_format_to_image_format() doesn't have ETC2 in its switch-case, so it returns VK_FORMAT_UNDEFINED, and then get_ahb_buffer_format_properties2() just continues without checking and ends up calling AHardwareBuffer_getNativeHandle with a bad handle. Mesa already has the ETC2 emulation infrastructure in vk_texcompress_etc2.h (similar to what they do for ASTC), it's just not wired up in ANV. I was thinking about submitting a patch upstream for that too, but Mesa ANV review cycles tend to take a while so this layer would still be useful as an immediate fix in the meantime.

Let me know how you'd like me to proceed on the dispatch table.

aleasto: Because this will be hooked into every vulkan user, correctness is a must have. So yes, please handle multiple instances.

uzairdeveloper223: Ok, I will work on it after some time.

aleasto: BTW we already track mesa with a few patches so you may also propose a mesa patch. https://github.com/waydroid/android_external_mesa3d

uzairdeveloper223: > BTW we already track mesa with a few patches so you may also propose a mesa patch. https://github.com/waydroid/android_external_mesa3d


Oh I already have made a patch for mesa upstream but I think I will apply directly to this. Thanks you for informing.

uzairdeveloper223: > BTW we already track mesa with a few patches so you may also propose a mesa patch. https://github.com/waydroid/android_external_mesa3d


So I took a look at this and there are only lineage 17 and 18 branches not 20 which I am currently working on but I will try to do something.

aleasto: That's because the lineage-18.1 branch also builds in lineage 20

uzairdeveloper223: Roger that!

uzairdeveloper223:
@aleasto, I also wrote a quick smoke test to verify the multi-instance dispatch on the host side. I can't build the full vendor image, so this was the easiest way to check correctness.

To run it yourself, first save and compile the test:

```c
// test_layer.c
#include <stdio.h>
#include <stdlib.h>
#include <vulkan/vulkan.h>

int main() {
    setenv("VK_LAYER_PATH", ".", 1);
    setenv("VK_INSTANCE_LAYERS", "VK_LAYER_WAYDROID_compat", 1);

    VkInstance inst1, inst2;
    VkInstanceCreateInfo ci = { .sType = VK_STRUCTURE_TYPE_INSTANCE_CREATE_INFO };

    printf("Creating instance 1...\n");
    if (vkCreateInstance(&ci, NULL, &inst1) != VK_SUCCESS) {
        printf("FAIL: vkCreateInstance 1\n"); return 1;
    }

    printf("Creating instance 2...\n");
    if (vkCreateInstance(&ci, NULL, &inst2) != VK_SUCCESS) {
        printf("FAIL: vkCreateInstance 2\n"); return 1;
    }

    uint32_t count = 0;
    vkEnumeratePhysicalDevices(inst1, &count, NULL);
    if (count == 0) { printf("No GPUs found\n"); return 1; }

    VkPhysicalDevice *devs = malloc(count * sizeof(VkPhysicalDevice));
    vkEnumeratePhysicalDevices(inst1, &count, devs);

    VkFormatProperties props;
    vkGetPhysicalDeviceFormatProperties(devs[0], VK_FORMAT_ETC2_R8G8B8_UNORM_BLOCK, &props);
    printf("ETC2 RGB8 features: linear=%u optimal=%u buffer=%u\n",
           props.linearTilingFeatures, props.optimalTilingFeatures, props.bufferFeatures);

    if (props.linearTilingFeatures == 0 && props.optimalTilingFeatures == 0)
        printf("PASS: ETC2 correctly masked\n");
    else
        printf("FAIL: ETC2 not masked\n");

    free(devs);
    vkDestroyInstance(inst2, NULL);
    vkDestroyInstance(inst1, NULL);
    printf("PASS: Both instances destroyed cleanly\n");
    return 0;
}
```

Then compile the layer and the test, temporarily changing the JSON library path to `./VkLayer_waydroid_compat.so` for local loading:

```bash
gcc -shared -fPIC -o VkLayer_waydroid_compat.so VkLayer_waydroid_compat.c -lvulkan -lpthread -fvisibility=hidden
gcc test_layer.c -o test_layer -lvulkan
./test_layer
```

Output on my machine (Intel UHD 620):

```
Creating instance 1...
Creating instance 2...
ETC2 RGB8 features: linear=0 optimal=0 buffer=0
PASS: ETC2 correctly masked
PASS: Both instances destroyed cleanly
```

Both instances are created and destroyed without issues, and ETC2 format features are correctly zeroed out.

aleasto:
Closing in favour of https://github.com/waydroid/android_external_mesa3d/pull/12

Thank you

PR #73, opened 2026-07-22
hwcomposer: Prevent null pointer dereference on layer buffer handle

herobuxx:
The hwcomposer (waydroid_mode_impl.h) and graphics com[poser crash intermittently when a layer or framebuffer target is processed with a null buffer handle.


aleasto:
I don't think this beaviour is allowed by the HWC 1 specification, so this will need some explaining...

PR #66, opened 2025-10-07
hwcomposer: Add touchpad pinch-to-zoom gesture support

supechicken:
Fixes https://github.com/waydroid/waydroid/issues/905
Fixes https://github.com/waydroid/waydroid/issues/1180
Partially fixes https://github.com/waydroid/waydroid/issues/1953

This works by mapping the touchpad pinch gesture to touch movements, which sounds much easier compared to patching AOSP sources :)

**However, when the cursor is close to the edge of window, this might not work due to the lack of space for mapping the gesture (can be mitigated by setting a lower value in `persist.waydroid.zoom_sensitivity`, but this also affects the sensitivity as well)**

The sensibility can be adjusted by these two props; a higher value means faster zoom-in/out speed when the gesture is performed, and vice versa (and more screen space is needed to perform the gesture):
- `persist.waydroid.zoom_sensitivity` (default: 120)

PR #79, opened 2026-07-31
hwcomposer: Translate trackpad scrolling into synthetic touch gestures

MichaelMKenny:
See comment at the top of #77

### Summary

Implements the feature requested in waydroid/waydroid#1180: two-finger
trackpad scrolling is translated into a synthetic touchscreen drag instead
of discrete mouse-wheel notches, so apps receive a real touch gesture —
1:1 finger tracking, native fling momentum from Android's own velocity
tracker, working pull-to-refresh, and natural-direction horizontal pagers
with snap-back. This is the same approach ChromeOS uses for Android apps
("input translation mode"), and the same synthetic-touch injection path as
PR #66 (pinch-to-zoom).

### How it works

- `wl_pointer` axis events with `axis_source` "finger" accumulate into a
  synthetic finger driven over the existing `INPUT_TOUCH` pipe: touch down
  at the cursor on gesture start, touch moves following the deltas at
  native event rate (timestamps preserved — this is what makes Android's
  velocity tracking correct), touch up on `axis_stop` (fingers lifted).
  Android's stock gesture machinery does everything else; no fling physics
  are reimplemented, and behavior is identical across toolkit versions.
- Physical mouse wheels (`axis_source` "wheel") keep the existing
  wheel-event behavior; compositors that never send `axis_source` also
  keep the old path.
- Hold gestures (`zwp_pointer_gestures_v1` v3 — libinput's designed signal
  for "fingers resting on the touchpad", motivated by exactly this use
  case) catch a fling in progress: two fingers resting shortly after a
  scroll inject a touch down that halts the fling, like touching a
  coasting list on a phone; moving continues the drag seamlessly, lifting
  without scrolling cancels so nothing is tapped. Ignored outside a short
  window after scrolling so idle finger-resting can never press UI.
- Edge handling: the synthetic finger anchors at the cursor inset from
  screen edges; a gesture that runs off screen is cancelled (palm
  sequence, no spurious fling) and re-pressed to continue. Clicks, pointer
  leave, or unexpected motion cancel cleanly.
- The touch pipe (and Android-side touch device) is now created in
  `finished_calibrating` instead of the first configure event — hosts
  without a touchscreen previously had no pipe until a window resize
  happened. It is no longer deleted/recreated on same-size configures, and
  the display-size properties are confirmed visible before the pipe
  appears so EventHub never sizes the device from unset properties. If the
  pipe has no reader yet, scrolling falls back to wheel events.

### Configuration

- `persist.waydroid.touch_scrolling` (default **enabled**) — re-read at
  every gesture start, so it can be toggled live for A/B comparison.
- `persist.waydroid.touch_scrolling_speed` — scroll-distance multiplier
  (tracking and fling scale together, as on a physical touchscreen).

### Note on the vendored protocol XML

`external/wayland-protocols` only carries pointer-gestures v2, which lacks
hold gestures, so the current upstream XML (v3) is vendored next to
`wayland-android.xml` with matching genrules. Bumping the
external/wayland-protocols snapshot instead would be cleaner long-term —
happy to rework if preferred. Compositors advertising < v3 simply don't
get the fling-catch behavior.

### Testing

Daily-driven on lineage-20 x86_64 under KWin (Plasma Wayland, 120Hz
trackpad laptop): 1:1 scrolling, flings, fling-catch, pull-to-refresh and
horizontal paging verified across Compose and View-based apps; long
multi-hour sessions with heavy scrolling are stable (with the fence-leak
PR applied).


supechicken:
FYI I have done a [similar thing](https://github.com/WayDroid-ATV/android_hardware_waydroid/commit/dd996cffed81ebfb3493a5ccf439c0fa160e1976) before (only for scrolling though), but your implementation looks better than me at first glance :)

MichaelMKenny:
Pushed a follow-up commit for a bug that showed up in daily use.

**Symptom:** resting two fingers on the trackpad — pausing to decide
whether to scroll — would sometimes tap, or long-press, whatever the
cursor happened to be over. Fully reproducible: rest two fingers, tilt
the fingertips very slightly, keep resting, lift.

**Cause:** the synthetic finger was pressed on the first finger-source
axis event, however small. Fingers settling on a touchpad do emit a few
tiny axis events, and a drag that short scrolls nothing in Android — but
the press was already down on whatever was under the cursor, so it became
a long press once the 500 ms timeout elapsed, and `axis_stop` then lifted
it as a normal touch up, i.e. a tap. The touch synthesis was faithful;
what was missing was Android's own precondition for treating a touch as a
drag at all.

**Fix**, three parts:

- Defer the touch down until the accumulated deltas would carry the
  finger past Android's touch slop (8 dp, converted through the same
  density `ViewConfiguration` uses). Below it the deltas are kept rather
  than consumed, so nothing is emitted at all and a genuine scroll still
  starts on its first move, coalesced. Real scrolls cross the slop within
  a couple of events, which is what a real touchscreen does too.
- Track each press's displacement from its own anchor and release it as a
  cancel, not a touch up, if it never passed the slop. This covers the two
  presses that cannot defer: the fling catch, which by design goes down
  without having moved, and the clutch re-press at a screen edge — a lift
  in the frame right after a clutch was also being reported as a tap.
- Arm the fling-catch window only when the lift could actually have
  started a fling: the finger still travelling at Android's minimum fling
  velocity, and no pause since its last movement. A lift after the fingers
  had already settled now leaves nothing for a later hold to catch. The
  window itself is shortened to the span a fling can plausibly still be
  coasting, and is cleared when a gesture is cancelled or the touch pipe is
  recreated.

The third point is what actually makes the PR description's claim about
idle finger-resting true — before this, the hold catch fired after any
recent lift, whether or not a fling had been started.

**Two behavior changes worth knowing about:** a sub-slop nudge scroll now
does nothing at all (previously it did not scroll either, but did tap),
and lifting within one frame of an edge clutch loses that one fling. Both
are deliberate — never tapping seemed clearly better than never losing a
fling.

Verified on the same setup as the original PR (lineage-20 x86_64, KWin
Plasma Wayland, 120 Hz trackpad laptop): the repro above no longer
produces any touch events, and scrolling, flings, fling-catch,
pull-to-refresh and horizontal paging are unchanged.

Kept as a separate commit so the fix is reviewable on its own — happy to
squash it into the feature commit if you would rather the PR read as one
change.

PR #80, opened 2026-09-01
audio: honour the format and rate AudioFlinger requests

kanya-approve:
## The problem

`adev_open_output_stream()` hardcodes every output stream to `S16_LE` at
`PLAYBACK_CODEC_SAMPLING_RATE`:

```c
out->config.rate = PLAYBACK_CODEC_SAMPLING_RATE;
out->config.format = SND_PCM_FORMAT_S16_LE;
```

and `pcm_format_from_audio_format()` treats float as unrepresentable:

```c
case AUDIO_FORMAT_PCM_FLOAT:  /* there is no equivalent for float */
default:
    LOG_ALWAYS_FATAL(...)
```

ALSA does have an equivalent — `SND_PCM_FORMAT_FLOAT_LE`/`_BE`. Because of the
`LOG_ALWAYS_FATAL`, the format could not even be *tested* against the table
without aborting audioserver, so the comparison in `adev_open_output_stream()`
had to be arranged to never pass float in.

The effect is that everything the container plays is resampled to one rate and
truncated to 16-bit before it reaches the host. 44.1kHz content is resampled to
48kHz just as hi-res is.

The renegotiation branch does not rescue this. It sets `ret = -EINVAL` and
writes a corrected config back, but the function ends in `return 0`, so the
error is discarded (pre-existing, untouched by this PR). Even if it were
returned, `AudioHwDevice::openOutputStream()` calls the HAL once and carries a
`FIXME` acknowledging it ignores the config that comes back. There is no retry,
so the first call has to be the one that matches.

## The change

Two commits:

**1. Map `AUDIO_FORMAT_PCM_FLOAT` to `SND_PCM_FORMAT_FLOAT_*`.** Adds the
missing cases in both directions. The conversion is split into
`pcm_format_from_audio_format_safe()`, which returns
`SND_PCM_FORMAT_UNKNOWN` for anything unconvertible, and the original
`pcm_format_from_audio_format()`, now a wrapper that keeps the fatal contract
for existing callers — `adev_open_input_stream()` still aborts exactly as
before. Nothing that relied on the old behaviour changes.

**2. Open output streams at the requested format and rate.** Uses the probe to
take AudioFlinger's format when it is one the table knows, falling back to
`S16_LE` otherwise, and takes its rate unless it is zero. `snd_pcm_hw_params`
still negotiates, so an unsupported rate snaps to the nearest the backend
offers rather than failing.

The policy XML widens `primary output` to advertise float and the standard
family of rates, since the HAL can now honour them.

Deliberately not included: a `DIRECT` mix port. I tried one first, assuming
hi-res needed to bypass the mixer. It is not necessary — the primary output
carries 192kHz once the HAL stops overriding the request — so it is not here.

## Testing

x86_64, Waydroid on lineage-20, an iFi USB DAC via PipeWire, playing 24/192
content.

Before, at the host ALSA device:

```
format: S16_LE
rate:   48000
```

After:

```
format: S32_LE
rate:   192000 (192000/1)
Momentary freq = 192002 Hz (0x18.0010)
```

The USB feedback value is the endpoint's own clock reporting lock at the
requested rate, so this is the hardware confirming it, not a software claim.
44.1kHz content now also arrives as 44100 rather than being resampled to 48000.

## What I have not tested

- **Big-endian.** The `_BE` cases mirror the existing `_BE` entries around them
  but were never executed.
- **ARM.** Built and run only on x86_64.
- **Input streams.** Untouched, and `adev_open_input_stream()` keeps the fatal
  converter.
- **Devices that report a narrow rate set.** Negotiation should snap to the
  nearest supported rate, but I only have hardware that accepts the full range.

One consequence worth flagging: advertising float means AudioFlinger stops
converting to 16-bit itself and hands float through, so the mixer output format
genuinely changes rather than merely widening. On a backend that cannot do
float, `snd_pcm_hw_params` refuses and the stream fails to open — which is the
case I cannot exercise here and would most want review on.

### DIFF
--- a/audio/audio_hw.c
+++ b/audio/audio_hw.c
@@ -286,13 +286,14 @@ static audio_channel_mask_t out_get_channels(const struct audio_stream *stream)
     return audio_channel_out_mask_from_count(out->config.channels);
 }
 
-/* Converts audio_format to pcm_format.
+/* Converts audio_format to pcm_format without aborting.
  * Parameters:
  *  format  the audio_format_t to convert
  *
- * Logs a fatal error if format is not a valid convertible audio_format_t.
+ * Returns SND_PCM_FORMAT_UNKNOWN if format is not a valid convertible
+ * audio_format_t, so a caller can probe a format before committing to it.
  */
-static inline snd_pcm_format_t pcm_format_from_audio_format(audio_format_t format)
+static inline snd_pcm_format_t pcm_format_from_audio_format_safe(audio_format_t format)
 {
     switch (format) {
 #if HAVE_BIG_ENDIAN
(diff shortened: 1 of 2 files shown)


kanya-approve:
@aleasto

kanya-approve:
@erfanoabdi @aleasto when you get a chance, could you take a look at this one? Happy to rework anything.

PR #81, opened 2026-09-15
Some hwcomposer fixes/workarounds.

krnlyng:


hwcomposer/wayland-hwc.cpp
```diff
@@ -498,12 +522,13 @@ window::create(struct display *display, bool use_subsurfaces, std::string appID,
     window->set_app_id(std::move(appID));
 
     wl_surface_commit(window->surface);
-    // Wait for first configure event
-    do {
-        wl_display_roundtrip(display->display);
-    } while (!window->configured);
```
aleasto, line 504: mmh I recall this being a fragile piece of logic. could you try arguing why it is safe to remove for all windows after calibration?

krnlyng: The problem this PR addresses is that wl_display_dispatch and wl_dispatch_roundtrip cannot be executed from different threads without issues according to wayland docs. each thread should have their own queue. or some other wayland functions need to be used like wl_display_prepare_read etc.

This particular change is part of that, it makes sure we only call wl_display_roundtrip before the hwc_wayland_thread is created, but never while the thread is already running.

But then again, looking at this again, it might cause problems for multi-window scenarios of waydroid but that is something i haven't tested.

aleasto: mmh because we're not being called from the thread spinning the wayland mainloop here, I see...

krnlyng: Btw the issue this fixes, or works around is that when a window gets opened hwc might deadlock here.

krnlyng: And even if this is called from the same thread as the wayland mainloop, it should not happen recursively either of course. And that includes indirect recursion.

aleasto: Right. Ugh.

krnlyng: I'm not sure but a simple mutex lock around both invocations might help too. Not sure though.

krnlyng: But it could also cause configure events (?) or registry events (?) to be missed or handled incorrectly iirc.

